Issue #7, opened 2015-11-26
Classes probably shouldn't be in org.apache.spark.* and org.apache.hadoop.* namespaces


rxin:
If this is library going to exist as a package outside of Spark then its classes should probably not be under the org.apache.spark and org.apache.hadoop namespaces. Making these changes would make it easier to spot where this library depends on internal Spark + Hadoop functionality.

Link from https://github.com/HyukjinKwon/spark-xml/issues/4

JoshRosen:
Fixed in #9.

Issue #1, opened 2015-11-26
Scoverage and Codecov integration


rxin:
Linked from https://github.com/HyukjinKwon/spark-xml/issues/10

It would be nice to have Scoverage and Codecov integration for test coverage report. As what spark-avro does. See [here](https://github.com/databricks/spark-avro/blob/master/build.sbt#L47-L50) and the "Codecov" label in the README file.

PR #9, opened 2015-11-27
Classes probably shouldn't be in org.apache.spark.* and org.apache.hadoop.* namespaces

HyukjinKwon:
https://github.com/databricks/spark-xml/issues/7

1. I just renamed the packages from `org.apache.*` to `com.databricks.*`.

2. Polished imports and reordered them.

3. Updated some accessors to stay independent.

4. Just removed `coverage` which currently does not anything at `.travis.yml`

5. At `InferSchema`, removed `DecimalType` check (it is okay because this package does not infer `DecimalType`), so that it can be independent as`IntergrelType` is not used anymore which has an accessor for only `sql` package in Spark.


HyukjinKwon:
@JoshRosen Would you look though this please?

src/main/resources/META-INF/services/org.apache.spark.sql.sources.DataSourceRegister
```diff
@@ -1 +1 @@
-org.apache.spark.sql.xml.DefaultSource15
\ No newline at end of file
+com.databricks.spark.xml.DefaultSource15
```
HyukjinKwon: I already made some branches to open PRs after this PR.
In my local branch (for adding a short name test), I added a newline here.

src/main/java/com/databricks/hadoop/mapreduce/lib/input/XmlInputFormat.java
```diff
@@ -29,10 +30,11 @@
 import org.apache.hadoop.mapreduce.InputSplit;
 import org.apache.hadoop.mapreduce.RecordReader;
 import org.apache.hadoop.mapreduce.TaskAttemptContext;
+import org.apache.hadoop.mapreduce.lib.input.FileSplit;
+import org.apache.hadoop.mapreduce.lib.input.TextInputFormat;
```
HyukjinKwon: In alphabetic ordering, this `org.apache.hadoop.mapreduce.lib` should go above but I will update them when downgrading the API version in another PR for Hadoop version compatibility, which I will open right after closing this.

.travis.yml
```diff
@@ -27,7 +27,7 @@ matrix:
       scala: 2.11.7
       env: TEST_SPARK_VERSION="1.5.0"
 script:
-  - sbt -Dspark.testVersion=$TEST_SPARK_VERSION ++$TRAVIS_SCALA_VERSION coverage test
```
JoshRosen: Why remove this?

JoshRosen: Even if Codecov isn't set up, this should still end up printing coverage metrics to the console, so there's no harm in leaving it in.

HyukjinKwon: I just wanted to set up this later and properly for this issue https://github.com/databricks/spark-xml/issues/1.

HyukjinKwon: Can I get that back at another PR for that? I will open them right after closing this.

JoshRosen: Sure. Let me go ahead and merge this now. FYI, we should use the `./dev/merge_pr.py` script to merge pull requests and _not_ the GitHub merge button (similar to what we do in our other repos and in Spark itself).

JoshRosen:
Merged. Thanks!

.travis.yml
HyukjinKwon: Thanks, I also definitely will.

PR #13, opened 2015-12-01
Update license, document and organization

HyukjinKwon:
In this PR, I changed the license, the document and organizations written at `build.sbt`.
Probably I should have made each PR separately but I just did this as there are no other PRs except mine currently.

1. As the package name has been renamed, I updated the document and removed **Using with Spark shell** instruction because that would not work as the examples below in README.md due to the package name. I will get this back as soon as I make this work properly and check if this package is available via `--packages`. 

2. Updated license comments.

3. Changed the organizations at `build.sbt`

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,7 +103,7 @@ df.select("author", "id").collect().foreach(println)
 
 You can manually specify the schema when reading data:
 ```scala
-import of.SQLContext
+import org.apache.spark.sql.SQLContext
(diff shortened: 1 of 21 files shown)


HyukjinKwon:
Since I changed comments and several configurations in `build.sbt`, there would not be any harm to this library itself.

Due to the issue such as https://github.com/databricks/spark-xml/issues/16, I will merge this. I will correct some if there are something wrong.

HyukjinKwon:
Merged.

Issue #6, opened 2015-11-26
Should test against Hadoop 1.x and Hadoop 2.x in Travis


rxin:
This library's tests should be run against both Hadoop 1.x and 2.x in Jenkins, since binary incompatibilities in Hadoop's TaskAttemptContext class may mean that your published artifact may not be compatible with both versions. See https://github.com/databricks/spark-avro/pull/79 for an example of a similar PR against spark-avro and https://github.com/harsha2010/magellan/pull/25 for a PR against magellan, another Spark package.

I think this is especially important for this library, given that it includes a Hadoop InputFormat.

Link from https://github.com/HyukjinKwon/spark-xml/issues/5

HyukjinKwon:
@JoshRosen I changed (locally) the `TextInputFormat` from Hadoop 2.x (`org.apache.hadoop.mapreduce`) to 1.x (`org.apache.hadoop.mapred`) and I checked the codes run okay on Hadoop 2.x.

As it looks Hadoop supports full binary compatibility (according to [here](https://hadoop.apache.org/docs/current/hadoop-project-dist/hadoop-common/Compatibility.html) and [here](http://hortonworks.com/blog/running-existing-applications-on-hadoop-2-yarn)), I think I better use old APIs for this and then does not add separate tests for that. (I am somehow a bit worried though). 

Do you think this would be okay and safe?

JoshRosen:
Independently of how we change the actual code, it's important that our automated tests run all of our tests against both Hadoop 1.x and 2.x, since we also need to guard against future changes which might introduce incompatibilities. We don't necessarily need _dedicated_ tests for Hadoop compatibility, though; just running the existing tests with Hadoop versions configured in the builds (like I did in the issues linked above) should probably be sufficient for now.

HyukjinKwon:
I see. Thanks!

HyukjinKwon:
Fixed in https://github.com/databricks/spark-xml/pull/14

PR #14, opened 2015-12-01
Should test against Hadoop 1.x and Hadoop 2.x in Travis

HyukjinKwon:
https://github.com/databricks/spark-xml/issues/6

In this PR, I added the test metrics for Hadoop version compatibility.

Also, I simply downgraded the API version from Hadoop 2.x (`org.apache.hadoop.mapreduce`) to 1.x (`org.apache.hadoop.mapred`) as it looks Hadoop supports full binary compatibility (according to [here](https://hadoop.apache.org/docs/current/hadoop-project-dist/hadoop-common/Compatibility.html) and [here](http://hortonworks.com/blog/running-existing-applications-on-hadoop-2-yarn)),

In addition, I renamed the package according to the Hadoop 1.x package name.


HyukjinKwon:
I thought I should use `hadoopFile` for Hadoop 1.x API as I saw `TextFile` in csv datasource. I will try to test that in capacity myself (and also correct arguments).

HyukjinKwon:
Hm.. for me in terms of progress reporting and IO metrics at `HadoopRDD` and `NewHadoopRDD`, it think there is no differences (maybe I have missed something). Though, I updated this back for `newAPIHadoopFile` locally.

For me, I think it might not have to switch it to `newAPIHadoopFile` by reflecting a method but for the future differences (or maybe existing things I missed), this would be okay in a way. No strong preference.

I will add a commit for `newAPIHadoopFile` as soon as you say that would be better.

HyukjinKwon:
Although it looks there are not considerable advantages in terms of progress reporting and IO metrics, I decided to stay in Hadoop 2.x because both there can be still an advantage for the future update and, other datasources in Databricks and the JSON datasource also follow this way which I do not have to avoid by another way.

So I will add a commit for this.

HyukjinKwon:
Like the same reason, I might have to convert the Java class into Scala one but I will do it later in another PR.

HyukjinKwon:
I will merge this and open another PR soon for converting this into Scala (or to polish nits if there are) as anyway this added the version compatibility tests.

Issue #8, opened 2015-11-26
XML file Write support


rxin:
For now, this only supports to read XML files.

Now I left this open since it is tricky to convert to proper XML files.

Link from https://github.com/HyukjinKwon/spark-xml/issues/3

HyukjinKwon:
Fixed in https://github.com/databricks/spark-xml/pull/15

Issue #3, opened 2015-11-26
Using SAX instead of DOM for type inference and Row conversion


rxin:
Building a full DOM tree for each XML record is very expensive and unnecessary, since we only need to traverse the contents once for either type inference or Row conversion. Using SAX can probably win much better performance.

Link from https://github.com/HyukjinKwon/spark-xml/issues/8

HyukjinKwon:
For this issue, I got a reason to use DOM. Mostly the main reason is to infer ArrayType. In order to infer ArrayType, it needs to check all the child keys (child elements) in each record in a XML and decide it is a StructType when it has a single key or ArrayType when it has duplicated multiple keys.

(I also manually gave a precedence for ArrayType higher than StructType at InferSchema in case that some records have a single key (field) A, namely inferred as StructType, while others can have duplicated multiple keys (fields) A, namely inferred as ArrayType, where key (field) A should be inferred as ArrayType)

At the first I used StAX, but I realised that the implementation actually works almost like DOM (It just read all the elements for a record first and then started to process). So I changed it to DOM #2.

Maybe there is another clever way but I could not come up with one. It would be great if there is a good way to resolve the problem with SAX.

I think anyway it can use SAX for constructing Rows as the problem is only about type inference.

Link from https://github.com/HyukjinKwon/spark-xml/issues/8#issuecomment-159764517

HyukjinKwon:
Ah. Actually I think I can do this reducing `ObjectType` having the smae fields to `ArrayType` after inferring types for each. I will try.

HyukjinKwon:
Fixed in https://github.com/databricks/spark-xml/pull/43.

HyukjinKwon:
I used StAX parser instead of SAX because I can easily resample JSON datasource in Spark which uses Jackson parser (pull model). Since the issue was not to build a full DOM tree but to traverse the contents once, I anyway closed this issue.

PR #15, opened 2015-12-01
XML file Write support

HyukjinKwon:
https://github.com/databricks/spark-xml/issues/8
In this PR, I added a write support. Just like CSV datasource, it only supports to overwrite or write a new file. In more detail,

1. I added the write support by generating XML documents with `StaxXmlGenerator` just like `JacksonGenerator` at Spark 1.4.

2. added some tests similarly with CSV datasource.

3. renamed `rootTag` to `rowTag` and added `rootTag` for the option as writing a XML file. `rootTag` means the actual root of a XML file and `rowTag` is the tag of its sub-element.

4. let the `rootTag` and `rowTag` be optional (default values are `ROW` and `ROWS` individually).

Here I have several things to mention.

1. I used StAX instead of DOM. As mentioned in https://github.com/databricks/spark-xml/issues/3#issuecomment-159824785, there is no reason to use DOM parser except type inference.

2.  When overwriting a XML file without any options, it sets `ROW` and `ROWS` for `rootTag` and `rowTag` as options for writing as default. This does not affect the schema or data itself but the actual file contents (the root element and its sub-element names) . I wonder if it is okay to let it allows this.

3. The option key for `rootTag` is changed as I said above. I did not want to change the user-interface but decided to do that because the meaning of `rootTag` may be misunderstood as actual root of a XML file, and any release is not published yet for this repository.


HyukjinKwon:
I unintendedly changed a lot of codes. I will take a look at the codes closely again.

src/main/scala/com/databricks/spark/xml/package.scala
```diff
@@ -48,11 +54,59 @@ package object xml {
...
+
+        new Iterator[String] {
+          var headingTag: Boolean = true
+          var trailingTag: Boolean = true
+          override def hasNext: Boolean = iter.hasNext || headingTag || trailingTag
+
+          override def next: String = {
+            if (iter.nonEmpty) {
+              val writer = new StringWriter()
+              val xmlWriter = factory.createXMLStreamWriter(writer)
```
HyukjinKwon: I will let this reuse writer rather than re-creating it for every row..

src/main/scala/com/databricks/spark/xml/parsers/stax/StaxXmlGenerator.scala
```diff
@@ -0,0 +1,82 @@
...
+      case (DateType, v) => writer.writeCharacters(v.toString)
+      case (udt: UserDefinedType[_], v) => writeElement(udt.sqlType, udt.serialize(v))
+
+      case (MapType(kv, vv, _), v: Map[_, _]) =>
+        v.foreach { p =>
+          writer.writeStartElement(p._1.toString)
+          writeElement(vv, p._2)
+          writer.writeEndElement()
+        }
+
```
HyukjinKwon: Roundtrip in reading and writing XML files is okay because It only can have the `ArrayType` with the structure `...ArrayType(StructType(ArrayType...))`. However, when it tries to write a just normal `Dataframe` having a structure `...ArrayType(ArrayType...)`, then it throws an exception.

I will add a test and fix this.

HyukjinKwon:
For the case roundtrip in reading and writing XML files, `ArrayType` cannot have `ArrayType` as element type. It always wraps the element of `ArrayType` with `StructType`. 
However,  when we convert a normal `DataFrame` to XML file. `ArrayType` can have `ArrayType` as elements. In this case, it is confusing what is this element name for XML file. Now, it is "item" but this might have to be changed (e.g. the same name with the parent field name).

src/main/scala/com/databricks/spark/xml/util/XmlFile.scala
```diff
@@ -25,13 +25,17 @@ import org.apache.spark.rdd.RDD
 import com.databricks.hadoop.mapreduce.lib.input.XmlInputFormat
 
 private[xml] object XmlFile {
+  val DEFAULT_ROW_TAG = "ROW"
+  val DEFAULT_ROOT_TAG = "ROWS"
+  val DEFAULT_INDENT = "    "
   val DEFAULT_CHARSET = Charset.forName("UTF-8")
+  val DEFAULT_ROW_SEQ = "\n"
```
HyukjinKwon: It is confusing if this is the right place for these.

HyukjinKwon:
@JoshRosen Would you mind if you quickly look through this?

I am not too sure if this approach and codes are okay, in particular, adding an element "item" for the element in `ArrayType` of `ArrayType`.

HyukjinKwon:
I am going to merge this as it anyway needs to support to write.

Issue #10, opened 2015-11-30
apply in an companion object probably should return the object of the companion class.


HyukjinKwon:
According to [Databricks Scala style](https://github.com/HyukjinKwon/scala-style-guide/blob/master/README.md#apply_method),  `apply()` of an companion object would better just return the object of the  companion class.

`DomXmlPartialSchemaParser` and `DomXmlParser` are not doing this.
For `InferSchema`, probably it also should not use `apply()`.

HyukjinKwon:
Fixed in https://github.com/databricks/spark-xml/pull/19

PR #19, opened 2015-12-07
apply in an companion object probably should return the object of the companion class.

HyukjinKwon:
https://github.com/databricks/spark-xml/issues/10

According to [Databricks Scala style](https://github.com/HyukjinKwon/scala-style-guide/blob/master/README.md#apply_method),  `apply()` of an companion object would better just return the object of the  companion class. However, the objects of `InferSchema`, `DomXmlParser` and `DomXmlPartialSchemaParser` are not doing this.

I changed the function names correspondingly with Spark JSON datasource (`InferSchema` and `JacksonParser`).


HyukjinKwon:
I will merge this since this PR changed the function names correspondingly to Spark, which I do not think we need to discuss on this.

HyukjinKwon:
Merged..

PR #20, opened 2015-12-07
Add some tests for type casting with signs.

HyukjinKwon:
The tests for sign-safety are added in this PR.


HyukjinKwon:
I will merge this since it is only about testing which does not harm the library.

HyukjinKwon:
Merged.

Issue #17, opened 2015-12-07
An option for parsing mode such as PERMISSIVE, DROPMALFORMED and FAILFAST


HyukjinKwon:
Sometimes for users, It should be permissive of rows that have fewer tokens than the schema. Also, it needs to support various paring modes like https://github.com/databricks/spark-csv/pull/37.

HyukjinKwon:
Since XML files can have different schemas, it might not need PERMISSIVE mode. I will open a PR for this dealing with only DROPMALFORMED and FAILFAST as boolean.

HyukjinKwon:
Fixed in #22

PR #22, opened 2015-12-08
An option for failing fast or dropping a malformed rows.

HyukjinKwon:
https://github.com/databricks/spark-xml/issues/17

As I mentioned in the issue, some paring modes should be supported for failing fast if it fails to parse or dropping the rows.

In this PR, PERMISIVE mode (mentioned in the issue) is excluded as XML itself can contains a different schema for each element.


src/main/scala/com/databricks/spark/xml/parsers/dom/DomXmlParser.scala
```diff
@@ -148,23 +154,32 @@ private[xml] object DomXmlParser {
...
+            None
+          }
           Some(convertObject(parser, schema))
+        } catch {
+          // Java library DOMParser throws SAXException when it fails.
+          case se: SAXException if !failFast =>
+            logger.warn(s"Dropping malformed row: ${xml.replaceAll("\n", "")}")
+            None
+          case se: SAXException if failFast =>
+            throw new RuntimeException(s"Malformed row (failing fast): ${xml.replaceAll("\n", "")}")
```
HyukjinKwon: I wonder if this  flatten xml `${xml.replaceAll("\n", "")}` would be a correct message form or not.

HyukjinKwon:
Maybe it should not use a boolean for this as it might have to support other modes. However, I cannot think of other modes for now.

PR #30, opened 2015-12-18
Use different indentation writer and not test with Hadoop 1.x and Spark 1.4.1.

HyukjinKwon:
`com.sun.xml.txw2.output.IndentingXMLStreamWriter` does not existing in some lower JDK. I changed this to `com.sun.xml.internal.txw2.output.IndentingXMLStreamWriter`

It only tests Spark 1.4.1 with Hadooop 2.2.0 due to the issue SPARK-8057.


HyukjinKwon:
Merged.

Issue #31, opened 2015-12-21
Java version compatibility


HyukjinKwon:
This library should take Java version compatibility between 6 & 7 JDK and 8 JDK into account due to `XmlInputFormat`.

I think it would be better to change it into Scala.

JoshRosen:
How is the choice of language (Java vs. Scala) related to JVM compatibility? These seem like orthogonal issues tom e.

HyukjinKwon:
Ah, I think I should correct the issue name and contents.

I just wanted to say that we might have to remove the Java code in order to not care of Java version stuff.

I complied this library compiled in Java 8 and I found I could not execute with Java 7 with the message below.

```
java.lang.UnsupportedClassVersionError: com/databricks/hadoop/mapreduce/lib/input/XmlInputFormat :
 Unsupported major.minor version 52.0
```

JoshRosen:
The presence of Java source code isn't the cause of that compatibility problem; rather, the issue is that the Java 8 compiler, by default, emits Java-8-compatible bytecode and needs to be explicitly instructed to generate bytecode which is compatible with Java 7. See https://github.com/databricks/spark-redshift/pull/113 for an example of how to configure `scalac` and `javac` to target specific JVM / JDK versions.

HyukjinKwon:
Thank you very much for your great explanation!

HyukjinKwon:
I will close this issue but leave the PR anyway because anyway I think it would be better to change Java code into Scala for now.

PR #32, opened 2015-12-21
Convert XmlInputFormat into Scala

HyukjinKwon:
https://github.com/databricks/spark-xml/issues/31

In this PR, Java code `XmlInputFormat` converted into Scala.


HyukjinKwon:
I am going to merge this as it passes all tests and is almost similar with an [`InputFormat`](https://github.com/databricks/spark-redshift/blob/master/src/main/scala/com/databricks/spark/redshift/RedshiftInputFormat.scala) in databricks/spark-redshift. So, functionally and structurally there would not be a big issue.

HyukjinKwon:
Merged.

PR #35, opened 2015-12-22
Polishing tests to remove useless variables and duplicated variables.

HyukjinKwon:
In this PR, I removed some useless newlines, variables and duplicated variables having the same value.

Now that the options `rowTag` and `rootTag` have the default values, `ROW` and `ROWS` in this library, I removed setting the options in some tests.

Also, the `rowTag`, `book`, can be shared instead of defining that for each file. 
It looks the functionality of `rowTag` and `rootTag` can be tested on some so it would be safe to remove others.


HyukjinKwon:
Merged.

PR #37, opened 2015-12-24
Add tests for types in writing and support for writing Decimal and MapType.

HyukjinKwon:
In this PR, I added some tests for writing with various types.
Also, I removed `BinaryType` which correctly does not work correctly and let an error occur when the type is not covered.


HyukjinKwon:
I think I should work around for `MapType` and remove `BinaryType`. I will submit more commits soon.

HyukjinKwon:
Actually, I am not too sure if `MapType` should be supported. I will open another PR for that.

HyukjinKwon:
Merged.

PR #38, opened 2015-12-28
Add test for malformed rows about types and some messages for them

HyukjinKwon:
In this PR, I added some warning messages when it fails to parse malformed rows about types.

Also, I added some tests for them. Additionally, I also changed some tests and the name of the test.


HyukjinKwon:
Merged.

Issue #21, opened 2015-12-07
Support for pruned scan for required fields.


HyukjinKwon:
Like https://github.com/databricks/spark-csv/pull/169, this library might also have to support pruned scan.
This would improve its performance in particular when the number of requested columns is few but the schema is huge according to the test I proceeded in the PR at CSV datasource above.

HyukjinKwon:
Fixed in https://github.com/databricks/spark-xml/pull/46.

PR #46, opened 2016-01-07
Support for pruned scan for required fields. 

HyukjinKwon:
https://github.com/databricks/spark-xml/issues/21
This library will read all data but does not have to check all the values in rows and convert (cast) them all.
This PR adds the feature for pruned scan for this.

For the reason similar with https://github.com/databricks/spark-csv/issues/219, this should parse all when dropping rows. This PR is similar with https://github.com/databricks/spark-csv/pull/169.

I did not add a test for this as it looks existing tests covers this.

### DIFF
--- a/src/main/scala/com/databricks/spark/xml/XmlRelation.scala
+++ b/src/main/scala/com/databricks/spark/xml/XmlRelation.scala
@@ -69,12 +69,33 @@ case class XmlRelation protected[spark] (
   }
 
   override def buildScan(requiredColumns: Array[String]): RDD[Row] = {
-    val requestedSchema = StructType(requiredColumns.map(schema(_)))
-    StaxXmlParser.parse(
-      baseRDD(),
-      schema,
-      requestedSchema,
-      parseConf)
+    val requiredFields = requiredColumns.map(schema(_))
+    val schemaFields = schema.fields
+    if (schemaFields.deep == requiredFields.deep) {
+      buildScan()
+    } else if (parseConf.failFastFlag) {
+      val safeRequestedSchema = StructType(requiredFields)
+      StaxXmlParser.parse(
+        baseRDD(),
+        safeRequestedSchema,
+        parseConf)
+    } else {
+      // If `parseConf` is disabled, then it needs to parse all the values
+      // so that we can decide which row is malformed.
+      val safeRequestedSchema = StructType(
+        requiredFields ++ schema.fields.filterNot(requiredFields.contains(_)))
+      val rows = StaxXmlParser.parse(
+        baseRDD(),
+        safeRequestedSchema,
+        parseConf)
+
+      val rowSize = requiredFields.length
+      rows.mapPartitions { iter =>
+        iter.flatMap { xml =>
+          Some(Row.fromSeq(xml.toSeq.take(rowSize)))
+        }
+      }
+    }
   }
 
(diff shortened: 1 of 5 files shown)


src/main/scala/com/databricks/spark/xml/XmlRelation.scala
```diff
@@ -67,6 +68,27 @@ case class XmlRelation protected[spark] (
...
+        // If `failFast` is disabled, then it needs to parse all the values
+        // so that we can drop malformed rows.
+        requiredFields ++ schemaFields.filterNot(requiredFields.contains(_))
+      } else {
+        requiredFields
+      }
+      StaxXmlParser.parse(
+        baseRDD(),
+        StructType(safeRequiredFields),
+        parseConf)
```
HyukjinKwon: When dropping rows with failing fast option disabled, I will return extra values as it needs to parse all the data to drop malformed rows. This will return correct results but I think I should slice the results before returning.

src/main/scala/com/databricks/spark/xml/XmlRelation.scala
```diff
@@ -60,13 +61,43 @@ case class XmlRelation protected[spark] (
...
+      // so that we can decide which row is malformed.
+      val safeRequestedSchema = StructType(
+        requiredFields ++ schema.fields.filterNot(requiredFields.contains(_)))
+      val rows = StaxXmlParser.parse(
+        baseRDD(),
+        safeRequestedSchema,
+        parseConf)
+
+      val rowSize = requiredFields.length
+      rows.mapPartitions { iter =>
```
HyukjinKwon: I am not too sure if it is okay to call `mapPartition` again to slice. I could do this in `StaxXmlParser.parse()` but I decided that this way looks much easy to read.

However, as this logic runs at most executions (by default, this logic would be executed), I am still puzzled if it is okay.

HyukjinKwon:
I tested the pure iteration time of produced RDD from `buildScan()` with `lineitem` table in TPC-H.
I loaded this by Spark CSV and wrote it to XML by this library.
On local machine, I set the number of 6,000 rows and measured the time.

The result was below. (The total number of columns was 17)

with `TableScan`,

|Scan Type|Required Fields|Elapsed Time|
|--------|-------|-------|
|`TableScan`|`C0`|894.05|
|`TableScan`|`C0, C1, C2, C3, C4, C5, C6`|918.36|
|`TableScan`|`* (C0 - C16)`|914.03|

with `failFast`, `true`,

|Scan Type|Required Fields|Elapsed Time|
|--------|-------|-------|
|`PrunedScan`|`C0`|724.42|
|`PrunedScan`|`C0, C1, C2, C3, C4, C5, C6`|784.67|
|`PrunedScan`|`* (C0 - C16)`|890.01|

with `failFast`, `false` (meaning dropping rows),

|Scan Type|Required Fields|Elapsed Time|
|--------|-------|-------|
|`PrunedScan`|`C0`|914.48
|`PrunedScan`|`C0, C1, C2, C3, C4, C5, C6`|923.34|
|`PrunedScan`|`* (C0 - C16)`|939.59|


While I do not see big regression in performance, it seems there is improvement in pruned scan with `failFast`, `true`. So, I am going to merge this for now.

HyukjinKwon:
Merged.

PR #56, opened 2016-01-13
Remove unreachable catches in type inference

HyukjinKwon:
This PR removes the unreachable catches since `NumberFormatException`, `ParseException` and `IllegalArgumentException` are raised when parsing invalid data into some types.


HyukjinKwon:
I am going to merge this.

PR #58, opened 2016-01-14
Fix a typo for version at example.

HyukjinKwon:
As the version is up, the version of example also should be updated.


HyukjinKwon:
Merged.

PR #60, opened 2016-01-14
Infer types correctly even if some elements has no attributes.

HyukjinKwon:
If some elements have no child but attributes and other do not, this ignores the attributes.
This PR fixes this by inferring types correctly.

- XML
```xml
   <book>
      <author>Gambardella, Matthew</author>
   </book>
   <book>
      <author id="bk102">Ralls, Kim</author>
   </book>
```

- Dataframe
```
root
 |-- author: string (nullable = true)
```


HyukjinKwon:
Merged.

PR #64, opened 2016-01-18
Fix to read attributes only in some elements and not to write attributes and elements for null

HyukjinKwon:
This PR fixes the bug while reading attributes. If there are attributes only in some elements in an array, it fails to read attributes. 

Also, this PR fixes this to write nothing for `null`.


HyukjinKwon:
Merged.

PR #70, opened 2016-01-22
Fix Python example in documentation

HyukjinKwon:
This PR fixes the Python example for including write just like the other examples.


HyukjinKwon:
Merged.

Issue #74, opened 2016-01-26
Support for namespaces declared at the ancestor level?


alreadyexists:
Hello! Some clarifications on applicability needed.
Suppose I'd like to extract `Topic` entries from DMOZ RDF dump (http://rdf.dmoz.org/)
WLOG, it goes like this:

``` xml
<?xml version="1.0" encoding="UTF-8"?>
<RDF xmlns:r="http://www.w3.org/TR/RDF/" xmlns:d="http://purl.org/dc/elements/1.0/" xmlns="http://dmoz.org/rdf/">
  <!-- Generated at 2016-01-24 00:05:51 EST from DMOZ 2.0 -->
  <Topic r:id="">
    <catid>1</catid>
  </Topic>
</RDF>
```

Topic attribute `id` refers to a namespace `r` declared in the RDF scope. If I'm getting the whole thing right, on Spark, when looking for `rowtag` entries within a particular partition, there is no way to tell that this namespaces have been defined in the outer scope, in a way, globally to the whole Spark cluster. This is why I'm getting 'Malformed row' warning in the log on this example (caused by `javax.xml.stream.XMLStreamException: ParseError at [row,col]:[1,16]
Message: http://www.w3.org/TR/1999/REC-xml-names-19990114#AttributePrefixUnbound?Topic&r:id&r`)

First of all, am I correct?
Second: any workaround here? I'd really enjoy reading that RDF dump. Maybe parser could expect for some static definitions provided by the user?
Thanks.

HyukjinKwon:
@alreadyexists  I see. Yes this library is not handling namespaces. Also, I think that sounds reasonable. Above all, I think this library should be able to read and parse the XML you just provided. I will look into this closer if I have some time.

HyukjinKwon:
I could reproduce this bug you meet by running below:

```scala
val testFile = "path-for-xml"
sqlContext.xmlFile(testFile, rowTag = "Topic").show()
```

The console output was 

```bash
11:25:32.517 WARN com.databricks.spark.xml.util.InferSchema$: Dropping malformed row: <Topic r:id="">        <catid>1</catid>    </Topic>
root
```

HyukjinKwon:
Just to make this clear, this was also suggested by @davemoyers, https://github.com/databricks/spark-xml/issues/39#issuecomment-167805138.

HyukjinKwon:
[...]

PR #75, opened 2016-01-27
Ignore namespaces while reading.

HyukjinKwon:
#74 

This is a workaround to read a XML files with namespaces.
Currently, this ignores namespaces but we might need to handle this by options or other ways.

This PR makes this library able to read a XML file rather than ignoring the rows as malformed rows below: 

```bash
11:25:32.517 WARN com.databricks.spark.xml.util.InferSchema$: Dropping malformed row: <Topic r:id="">        <catid>1</catid>    </Topic>
root
```


HyukjinKwon:
I am going to merge this.

HyukjinKwon:
Merged.

PR #78, opened 2016-01-29
Fix bugs while reading a field of mixed type elements due to structure conversion.

HyukjinKwon:
The XML is having mixed type of field , `date`. In this case, due to structure conversion, it sometimes does not set the value properly (it sets `null`s).

```xml
...
    <ROW>
        <date type="string">2012-12-01</date>
    </ROW>
    <ROW>
        <date type="struct">
            <year>2012</year>>
            <month>11</month>
            <day>2</day>
        </date>
    </ROW>
...
```

So, this PR fixes this bug to properly set the values.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,6 +1,6 @@
 name := "spark-xml"
 
-version := "0.3.1"
+version := "0.3.2-SNAPSHOT"
 
 organization := "com.databricks"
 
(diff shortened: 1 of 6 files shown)


HyukjinKwon:
I am going to merge this.

HyukjinKwon:
Merged.

Issue #80, opened 2016-02-07
Empty field causes scala.MatchError?


larsbkrogvig:
Hi,

I am having some issues reading a certain .xml file containing GPS data. 

The file I am trying to read contains a row that I suppose is malformed. I supply my own schema and use `failFast = false`. Instead of dropping the row, it fails. Is this intended? Or am I using the library incorrectly?

I use this schema:
```
root
 |-- @lat: double (nullable = true)
 |-- @lon: double (nullable = true)
 |-- ele: double (nullable = true)
 |-- extensions: struct (nullable = true)
 |    |-- gpxtpx:TrackPointExtension: struct (nullable = true)
 |    |    |-- gpxtpx:atemp: long (nullable = true)
 |    |    |-- gpxtpx:cad: long (nullable = true)
 |    |    |-- gpxtpx:hr: long (nullable = true)
 |-- time: string (nullable = true)
 |-- athlete: string (nullable = true)
 |-- activity_type: string (nullable = true)
```
I try to parse this file:
```
<trkseg>
 <trkpt lat="59.9155180" lon="10.6962430">
  <ele>0.0</ele>
  <time>2013-01-24T06:18:43Z</time>
  <extensions>
   <gpxtpx:TrackPointExtension>
   </gpxtpx:TrackPointExtension>
  </extensions>
 </trkpt>
</trkseg>
```
I get this error:
```
scala.MatchError: 
      (of class java.lang.String)
	at com.databricks.spark.xml.parsers.StaxXmlParser$$anonfun$com$databricks$spark$xml$parsers$StaxXmlParser$$convertObject$1.apply$mcVI$sp(StaxXmlParser.scala:218)
```
I think the problem may be caused by `<gpxtpx:TrackPointExtension>` being empty. Usually, it looks like this:
```
<gpxtpx:TrackPointExtension>
 <gpxtpx:hr>119</gpxtpx:hr>
 ...
</gpxtpx:TrackPointExtension>
```
In this case I'd perfer to have `gpxtpx:TrackPointExtension` be `null`. Alternatively, the row could be dropped. Is this a bug, or am I not seeing how I should do that?

Thanks in advance for taking the time.
Kind regards,
Lars

Bertrandbenj:
Hi,

I have a similar issue when I let spark-xml infer my schema. it infer a like 
<a>
 <b>something here</b>
</a>

but later in the document I have the same "a" tag containing more stuff 
<a>
 <test>some other node</test>
[...]

Issue #72, opened 2016-01-24
Datatype assumption causes document to be skipped.


davemoyers:
Hi,

I am testing with v0.3.1 pulling through Maven. I encountered an issue where it appears that the module is assuming a datatype of a nested element based upon the first occurrence, when in fact the datatype is something else. I am only able to reproduce it using this XML document: http://docs.oasis-open.org/ubl/os-UBL-2.1/xml/UBL-Order-2.1-Example.xml

When the document is processed I receive this error: WARN StaxXmlParser$: Number format exception. Dropping malformed line: <Order xmlns=...

Within the document there is a PartyIdentification section with an ID of "PartyID123": 

      <cac:PartyIdentification>
        <cbc:ID>PartyID123</cbc:ID>
      </cac:PartyIdentification>

If I change the ID to "123" instead of "PartyID123" the document is successfully processed. Alternatively, if instead of changing that element, I go to the prior PartyIdentification element and change its child ID element to include arbitrary text (notice the ABC appended to "7300070011115"):

      <cac:PartyIdentification>
        <cbc:ID schemeAgencyID="9" schemeID="GLN">7300070011115ABC</cbc:ID>
      </cac:PartyIdentification>
 
the document is then processed successfully. This is what makes me believe that some datatype is being determined/cached on the first occurrence, then attempted to be applied on subsequent occurrences. This behavior is not desirable because datatype should gracefully demote to a more generalized data type. For example if the first ID occurrence looked like a long type, but the second occurrence looks like a string, then the overall datatype should be assumed to be a string, not a long.

Correcting the datatype assumption will be important for more complex XML documents that do not reference a schema.

Thanks,
Dave

HyukjinKwon:
@davemoyers Thanks for reporting this and looking into this!
I will also try to investigate further.

HyukjinKwon:
It looks this issue is related with https://github.com/databricks/spark-xml/issues/80.
Also, I created a PR for this here, https://github.com/databricks/spark-xml/pull/82.

HyukjinKwon:
@davemoyers  Would you please try this with 0.3.2? I just released a new one. I could reproduce this and I fixed that.

davemoyers:
Hi Hyukjin,

I executed against 0.3.2 and my unit test failed with the same error. Here
are a few observations:

1. FYI: Although Maven Central shows 0.3.2, my project could not pull it
down, and I was forced to manually download the jar and drop it into my
[...]

PR #82, opened 2016-02-11
Create the correct type for the field having both an empty value and struct value during type inference.

HyukjinKwon:
https://github.com/databricks/spark-xml/issues/80 https://github.com/databricks/spark-xml/issues/72

As it is described, this library fails to infer type having  both an empty value and struct value.

Since it is usual for XML, this PR fixes the issue to infer type correctly.


HyukjinKwon:
I am going to merge this for now..

Issue #84, opened 2016-02-12
Slow performance for reading XML due to the XML library and not optimized logics.


HyukjinKwon:
Sameer Farooqui said:

> I'm reading the wikipedia file like this:
> 
> ```scala
> val wiki001DF = sqlContext.read
>     .format("com.databricks.spark.xml")
>     .option("rowTag", "page")
>     .load("/mnt/wikipedia-readonly/en_wikipedia/enwiki-20160204-pages-articles-multistream.xml")
> ```
> 
> When reading the 54 GB XML file, I noticed that on a Spark cluster with 15 Executors (4 cores on each), it took 40 minutes to read the file (about 839 tasks). Each task is running for about 2 minutes. I found that very slow/long.
> 
> I then took some thread dumps on the Executors and noticed that a lot of time is being spent in the xerces XML parsing library:
> 
> ```
> com.sun.org.apache.xerces.internal.impl.XMLEntityScanner.scanName(XMLEntityScanner.java:716)
> com.sun.org.apache.xerces.internal.impl.XMLDocumentFragmentScannerImpl.scanEntityReference(XMLDocumentFragmentScannerImpl.java:1844)
> com.sun.org.apache.xerces.internal.impl.XMLDocumentFragmentScannerImpl$FragmentContentDriver.next(XMLDocumentFragmentScannerImpl.java:3067)
> com.sun.org.apache.xerces.internal.impl.XMLDocumentScannerImpl.next(XMLDocumentScannerImpl.java:606)
> com.sun.org.apache.xerces.internal.impl.XMLNSDocumentScannerImpl.next(XMLNSDocumentScannerlmpl.java:117)
> com.sun.org.apache.xerces.internal.impl.XMLStreamReaderImpl.next(XMLStreamReaderImpl.java:558)
> com.sun.xml.internal.stream.XMLEventReaderImpl.peek(XMLEventReaderImpl.java:276)
> com.databricks.spark.xml.parsers.StaxXmlParserUtils$.readDataFully(StaxXmlParserUtils.scala:29)
> ```

Although this issue sounds a bit broad, I think I should create this as a remainder. It looks there are two things to deal with.

- Better XML library.
- Optimization for parsing logic.

HyukjinKwon:
Let me close this for now because as I said it is a bit broad and also the PR I submitted for this was pretty much successful in terms of performance improvement.

Green-Angry-Bird:
[...]

PR #85, opened 2016-02-12
XML factory per partition instead of each row and reduce peek() call

HyukjinKwon:
https://github.com/databricks/spark-xml/issues/84

Firstly, the XML factory was being created for each row. Although this factory cannot be reused across the tasks because it is not serializable, it still can create the factory for each partition.

Secondly, It looks `peek()` is called multiple times for reading a complete data within an element. However, `IS_COALESCING` option for XML parser allows this to read only once.

Also, While looking through this library, I found some stylistic corrections and useless logics. I removed them.
1. I used `isIgnorableWhiteSpace()` so that it can differentiate some spaces between elements but it looks this function does not detect anything due to DTD is not accessible during parsing each row.
2. Corrected the indentations of parameters in function across multiple lines.
3. Removed duplicated logics.
4. Corrected `StaxXmlGenerator` to write attributes first always for safety (to prevent to try to write some data within an element first and then write attributes, which emits an exception)

### DIFF
--- a/src/main/scala/com/databricks/spark/xml/parsers/StaxXmlGenerator.scala
+++ b/src/main/scala/com/databricks/spark/xml/parsers/StaxXmlGenerator.scala
@@ -59,9 +59,12 @@ private[xml] object StaxXmlGenerator {
           }
         case _ if name.startsWith(options.attributePrefix) =>
           writer.writeAttribute(name.substring(options.attributePrefix.length), v.toString)
+
         // For ArrayType, we just need to write each as XML element.
         case (ArrayType(ty, _), v: Seq[_]) =>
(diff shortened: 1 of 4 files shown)


HyukjinKwon:
This PR make this library almost 6-9 times faster according to my test.

For a simple example, I tested some cases below with with the wiki dump 1G,

Original code:
```scala
df.printSchema()     // 7.3 min
df.count()           // 3.7 s
df.collect()         // 7.2 min
```

This PR:
```scala
df.printSchema()     // 1.2 min
df.count()           // 38 s
df.collect()         // 1.3 min
```

HyukjinKwon:
I am going to merge this.

HyukjinKwon:
Merged.

Issue #88, opened 2016-02-15
Values not being parsed correctly / showing as null 


spiu:
Hi,

I've noticed that some fields in an xml I'm parsing come up as `null`. I'm using the code from master (updated as of today) and Schema generation seems to be alright.

It seems to be a problem in `StaxXmlParser` sending an incorrect "sub-schema". But I could not get to the root of the problem yet.

Any ideas on what could be wrong?

[Sample file here](https://github.com/databricks/spark-xml/files/131037/fpml-sample.xml.txt)

```
<dtcc:genericProductDTCC id="Id_12345">
    <fpml:primaryAssetClass>Equity</fpml:primaryAssetClass>
    <fpml:productType productTypeScheme="urn:bank:product-type:isda-plus">Equity:Exotic:Option</fpml:productType>
    <fpml:productId productIdScheme="http://www.fpml.org/coding-scheme/product-taxonomy">Equity:Other</fpml:productId>
    <fpml:productId productIdScheme="http://www.dtcc.com/coding-scheme/Narrow-Broad-Id">Broad</fpml:productId>
    <fpml:embeddedOptionType>false</fpml:embeddedOptionType>
    <fpml:buyerPartyReference href="party1"></fpml:buyerPartyReference>
    <fpml:buyerAccountReference href="account1"></fpml:buyerAccountReference>
    <fpml:sellerPartyReference href="party2"></fpml:sellerPartyReference>
    <fpml:premium>
        <fpml:payerPartyReference href="party1"/>
        <fpml:payerAccountReference href="account1"/>
        <fpml:receiverPartyReference href="party2"/>
        <fpml:paymentAmount>
            <fpml:currency>USD</fpml:currency>
            <fpml:amount>0</fpml:amount>
        </fpml:paymentAmount>
        <fpml:paymentDate>
            <fpml:adjustableDate>
                <fpml:unadjustedDate>2014-10-28</fpml:unadjustedDate>
            </fpml:adjustableDate>
        </fpml:paymentDate>
    </fpml:premium>
    <fpml:effectiveDate>
        <fpml:unadjustedDate>2014-10-24</fpml:unadjustedDate>
    </fpml:effectiveDate>
    <fpml:expirationDate>
        <fpml:unadjustedDate>2016-10-21</fpml:unadjustedDate>
    </fpml:expirationDate>
    <fpml:underlyer>
        <fpml:basket>
            <fpml:openUnits>999</fpml:openUnits>
            <fpml:basketConstituent>
[...]

PR #89, opened 2016-02-16
Parse elements in array having attributes correctly

HyukjinKwon:
https://github.com/databricks/spark-xml/issues/88

If elements in array have some attributes, it parses and infers the data incorrectly. For example, the xml below:

```xml
<ROWSET>
    <ROW>
        <one attr="value">i'm one</one>
        <one attr="value">i'm one</one>
    </ROW>
</ROWSET>
```

produces the results below:

```
+----+
| one|
+----+
|null|
|null|
+----+
```

This was because of mistakes to deal with elements in `ArrayType`. This PR fixes this.

Also, I changed some relevant functions name to look relevant and removed a unused test file.


HyukjinKwon:
I added the https://github.com/databricks/spark-xml/pull/87 in here as well by mistake. Please ignore the changes in `README.md` and `build.sbt`

spiu:
I can confirm that this PR fixes the issue I raised

HyukjinKwon:
Merged.

Issue #94, opened 2016-02-21
Can't collect DataFrame with empty fields read with specified schema


larsbkrogvig:
I am having problems working with a DataFrame I loaded with a specified schema. 

The problem looks related to the fix for #80, and could be the same as #93 as it produces the same error in some cases. (I think it is [this one](https://issues.apache.org/jira/browse/SPARK-12624))

For the record I am using `pyspark`.

### Short version
I can't `collect()` a certain DataFrame with missing values in it that I have read with a specified schema, and i get `ArrayIndexOutOfBoundsException` when trying to do certain operations with it.

### Very long version
I have two files, File1 and File2:
File1
```
<row>
    <a>A</a>
    <b>
        <c>C</c>
    </b>
</row>
```
File2
```
<row>
    <a>A</a>
    <b>
    </b>
</row>
```
When I infer schemas from the two files separately I get this:
File1
```
root
 |-- a: string (nullable = true)
 |-- b: struct (nullable = true)
 |    |-- c: string (nullable = true)
```
and File2
```
root
 |-- a: string (nullable = true)
 |-- b: string (nullable = true)
```
As expected `b` is a `struct` in File1 and a `string` in File2. When I infer the schema from both files at the same time I get this schema:
```
root
 |-- a: string (nullable = true)
 |-- b: struct (nullable = true)
 |    |-- #VALUE: string (nullable = true)
 |    |-- c: string (nullable = true)
```
At first I was surprised to see the `#VALUE` field in `b`. After thinking about it I think it makes sense, since `b` is a `string` in File2 and if a string value exists there it must go somewhere. However, I'm not interested in the `#VALUE` field and I want to use the schema from File1 on both files. 

**This is where the problems start**

I save the first schema and specify it when reading both files again to a DataFrame `df`. I use `treatEmptyValuesAsNulls=True` and expect `c` to be `null` in the row from File2. When I try to use `df` I get errors.
The schema looks correct:
```
root
 |-- a: string (nullable = true)
 |-- b: struct (nullable = true)
 |    |-- c: string (nullable = true)
```
And I can have a look at the contents with `df.show()`:
```
+---+---+
|  a|  b|
+---+---+
|  A|[C]|
|  A| []|
+---+---+
```
This looks ok, but i would expect to see `[null]` and not `[]` in the second row. I think that is the cause of the following problem: Running `df.collect()` yields the nasty error:
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
[...]

Issue #93, opened 2016-02-19
Inconsistent element attributes


cjmatta:
I have a question about the behavior of this code when presented with a document such as this:
```xml
<library location="uptown">
    <shelf category="fiction">
        <book author="Lewis Carol">Alice's Adventures In Wonderland</book>
        <book author="James Joyce">Portrait of the Artist as a Young Man</book>
    </shelf>
    <shelf category="non-fiction">
        <book author="Carl Sagan">Billions and Billions</book>
        <book>Webster's Collegiate Dictionary</book>
    </shelf>
</library>
<library location="downtown">
    <shelf category="fiction">
        <book author="J.K. Rowling">Harry Potter and the Prisoner of Azkaban</book>
        <book author="Homer">The Iliad</book>
    </shelf>
    <shelf category="non-fiction">
        <book author="Stephen Hawking">A Brief History of Time</book>
        <book author="Benjamin Franklin">The Autobiography of Benjamin Franklin</book>
    </shelf>
</library>
```
As you can see there's a book in there without an `author` attribute.

I can load this xml and print the schema:
```
val test = sqlContext.read.format("com.databricks.spark.xml").option("rowTag", "library").load("maprfs:///user/cmatta/projects/test.xml")
test.printSchema

test: org.apache.spark.sql.DataFrame = [@location: string, shelf: array<struct<@category:string,book:array<struct<#VALUE:string,@author:string>>>>]
root
 |-- @location: string (nullable = true)
 |-- shelf: array (nullable = true)
 |    |-- element: struct (containsNull = true)
 |    |    |-- @category: string (nullable = true)
 |    |    |-- book: array (nullable = true)
 |    |    |    |-- element: struct (containsNull = true)
 |    |    |    |    |-- #VALUE: string (nullable = true)
 |    |    |    |    |-- @author: string (nullable = true)
```

If I want to select all the books by location and collect them I get a `java.lang.ArrayIndexOutOfBoundsException: 1` error. If I just collect the whole dataframe without select I can see that the "Webster's Collegiate Dictionary" book array only has one element, where I would have expected two: the value and a null representing the author. Is this a correct expectation?
```
test.collect
[...]

PR #95, opened 2016-02-25
Handling inconsistent element attributes

HyukjinKwon:
#93 #94

The size of the nested row for that case itself was different with the schema. So, when it accesses to the fields, it emits `java.lang.ArrayIndexOutOfBoundsException` exception as described in the issue above.

This PR sets `null` explicitly for the fields as well like the other field so that there is no inconsistent fields.


HyukjinKwon:
As @larsbkrogvig [confirmed](https://github.com/databricks/spark-xml/issues/94#issuecomment-189855254) this patch resolves the issue, I am going to merge this soon if there are not critical stuff to correct.

Actually, I have been a bit confused who I should cc to review my PRs. It would be great if anybody informs me.

HyukjinKwon:
Merged.

PR #96, opened 2016-02-25
Merge inferred types correctly for elements having no child but attributes

HyukjinKwon:
https://github.com/databricks/spark-xml/issues/72
This PR corrects type inference so that this can find a proper compatible type from two different types when the field corresponding with `valueTag`.

Currently, this library infers the schema having duplicated fields corresponding with `valueTag` when the types are different.
Since one of the original tests is related with this, I just correct the test and the data.


HyukjinKwon:
I am going to merge this.

Issue #117, opened 2016-04-29
No value in nested struct causes arrayIndexOutOfBounds


lokm01:
Hi,

I found a problem (not sure if this is more related to spark-xml or spark itself), but I was able to fix it with small change in the code of spark-xml.

The issue is with nested structs, namely when one of them is missing.

We have large XML documents similar to:
```xml
<?xml version="1.0" ?>
<root>
	<item>
		<a>a</a>
		<b>
			<c>c</c>
			<d>0.4</d>
			<es>
				<e>
					<ee>1</ee>
				</e>
				<e>
					<ee>2</ee>
					<eee>3</eee>
				</e>
			</es>
		</b>
		<f>f</f>
	</item>
		<item>
		<a>a</a>
		<b>
			<c>c</c>
			<d>0.4</d>
			<es></es>
		</b>
		<f>f</f>
	</item>
</root>
```
As you can see, the `<es>` element in the second record is null.

When doing any action on the dataframe, I get ArrayOutOfBoundsException.

    val df = sqlc.read
      .format("com.databricks.spark.xml")
      .option("rowTag", "item")
      .load("sample.xml")
    df.write.parquet("testout")  

Will produce:

> 16/04/29 12:22:49 ERROR DefaultWriterContainer: Aborting task.
java.lang.ArrayIndexOutOfBoundsException: 0
	at org.apache.spark.sql.catalyst.expressions.GenericInternalRow.genericGet(rows.scala:226)
	at org.apache.spark.sql.catalyst.expressions.BaseGenericInternalRow$class.getAs(rows.scala:34)
	at org.apache.spark.sql.catalyst.expressions.BaseGenericInternalRow$class.isNullAt(rows.scala:35)
	at org.apache.spark.sql.catalyst.expressions.GenericInternalRow.isNullAt(rows.scala:220)

From the stack trace, you see that the parquet writer is calling isNull on a non-existing item in the Row. I guess it assumes from schema that the Row is non-empty.
Spark-xml however produces an empty row in such a case.

I was able to fix this by building spark-xml locally and changing the StaxXmlParser.scala, namely the  convertRow function so that it checks whether the Row is empty and return null in such a case:

    val res = Row.fromSeq(row)
    if(res.size == 0) null else res

Would it be possible to fix this in the next release? It is kind of a blocker for us at the moment.

Thanks!

lokm01:
[...]

PR #118, opened 2016-04-29
Issue 117 - No value in nested struct causes arrayIndexOutOfBounds

lokm01:
https://github.com/databricks/spark-xml/issues/117

Added a check into convertRow function to return null instead
of empty Row object.

This will ensure that Spark doesn't throw ArrayIndexOutOfBounds
when there's no value in a nested struct.

Change-Id: Id796de2b861868543b7fc6e95a6cd56567be5d47
Signed-off-by: Jan Scherbaum <kmoj02@gmail.com>

### DIFF
--- a/src/main/scala/com/databricks/spark/xml/parsers/StaxXmlParser.scala
+++ b/src/main/scala/com/databricks/spark/xml/parsers/StaxXmlParser.scala
@@ -214,7 +214,7 @@ private[xml] object StaxXmlParser {
     // The fields are sorted so `TreeMap` is used.
     val convertedValuesMap = convertValues(valuesMap, schema)
     val row = TreeMap((fields ++ convertedValuesMap).toSeq : _*).values.toSeq
-    Row.fromSeq(row)
+    if(row.length == 0) null else Row.fromSeq(row)
   }
 
   /**
--- a/src/main/scala/com/databricks/spark/xml/parsers/StaxXmlParser.scala
+++ b/src/main/scala/com/databricks/spark/xml/parsers/StaxXmlParser.scala
@@ -214,7 +214,9 @@ private[xml] object StaxXmlParser {
     // The fields are sorted so `TreeMap` is used.
     val convertedValuesMap = convertValues(valuesMap, schema)
     val row = TreeMap((fields ++ convertedValuesMap).toSeq : _*).values.toSeq
-    if(row.length == 0) null else Row.fromSeq(row)
+    // Return null rather than empty row. For nested structs empty row causes
+    // ArrayOutOfBounds exceptions when executing an action
+    if(row.isEmpty) null else Row.fromSeq(row)
   }
 
   /**
--- a/src/test/resources/null-nested-struct.xml
+++ b/src/test/resources/null-nested-struct.xml
@@ -22,6 +22,7 @@
 		<b>
 			<c>c</c>
 			<d>0.4</d>
+			<!-- Issue 117 - This is where an empty Row would be produced instead of null -->
 			<es></es>
 		</b>
 		<f>f</f>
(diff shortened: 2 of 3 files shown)


HyukjinKwon:
@lokm01 Thanks for this PR and simplified examples in the issues. Let me try to reproduce the issue first. Meanwhile, it would be great if we have a test in `XmlSuite` for this.

lokm01:
@HyukjinKwon No problem, added the test.

src/main/scala/com/databricks/spark/xml/parsers/StaxXmlParser.scala
```diff
@@ -214,7 +214,7 @@ private[xml] object StaxXmlParser {
     // The fields are sorted so `TreeMap` is used.
     val convertedValuesMap = convertValues(valuesMap, schema)
     val row = TreeMap((fields ++ convertedValuesMap).toSeq : _*).values.toSeq
-    Row.fromSeq(row)
+    if(row.length == 0) null else Row.fromSeq(row)
```
HyukjinKwon: I think `if (row.isEmpty) null else Row.fromSeq(row)` might be better.
Also, could you maybe add a comment why this checking is necessary?

src/test/resources/null-nested-struct.xml
```diff
@@ -0,0 +1,29 @@
...
+	<item>
+		<a>a</a>
+		<b>
+			<c>c</c>
+			<d>0.4</d>
+			<es></es>
+		</b>
+		<f>f</f>
+	</item>
+</root>
```
HyukjinKwon: I could reproduce the exam exception with the data below:

```xml
<?xml version="1.0" ?>
<root>
  <item>
    <b>
      <es>
        <e>1</e>
      </es>
    </b>
  </item>
  <item>
    <b>
      <es></es>
    </b>
  </item>
</root>
```

(It might be even nicer if the data contains some comments within XML data.)

HyukjinKwon: (Don't forget to add a newline at the end of this file)

src/test/scala/com/databricks/spark/xml/XmlSuite.scala
```diff
@@ -18,10 +18,8 @@ package com.databricks.spark.xml
 import java.io.File
 import java.nio.charset.UnsupportedCharsetException
 import java.sql.{Date, Timestamp}
-
 import org.scalatest.{BeforeAndAfterAll, FunSuite}
 import org.apache.hadoop.io.compress.GzipCodec
-
```
HyukjinKwon: Actually, this ordering is fine. This is described in https://github.com/databricks/scala-style-guide#imports but I could not have some time to add a travis style check.

src/test/scala/com/databricks/spark/xml/XmlSuite.scala
```diff
@@ -657,4 +656,19 @@ class XmlSuite extends FunSuite with BeforeAndAfterAll {
 
     assert(results.size === numTopics)
   }
+
+  test("Missing nested struct") {
```
HyukjinKwon: (It would be nicer if we have more detailed title)

HyukjinKwon:
@lokm01 I could reproduce and test. So, I looked though Spark internal JSON and external Avro data source to check how they deal with. Generally, LGTM except for few comments. Let me double-check.

lokm01:
@HyukjinKwon Done, thanks for feedback

HyukjinKwon:
Thanks. This one is merged to master, https://github.com/databricks/spark-xml/commit/19eb2779cf232dc0239e54bf353fa609c8b05231.

Issue #125, opened 2016-05-18
Produces wrong order of columns for nested rows when user specifies a schema.


HyukjinKwon:
With the data here,

```xml
<ROWSET>
  <ROW>
    <c>
      <a>111</a>
      <b>222</b>
    </c>
  </ROW>
</ROWSET>
```

and with the codes below: 

```scala
val path = "path-to-your-file"
val policySchema = StructType(
  Seq(
    StructField("b", StringType, nullable = true),
    StructField("a", StringType, nullable = true)))
val schema = StructType(
  Seq(
    StructField("c", policySchema, true)))

val df = sqlContext.read
  .format("xml")
  .schema(schema)
  .load(path)

df.select("c.a", "c.b").show()
```

this produces the results below:

```bash
+---+---+
|  a|  b|
+---+---+
|222|111|
+---+---+
```

This is because it tries to set the value regardless of the index of schema for nested rows.

HyukjinKwon:
Fixed in #126

PR #126, opened 2016-05-20
Produces correct order of columns for nested rows when user specifies a schema

HyukjinKwon:
#125 

This PR produces correct order of columns when user specifies a schema.

Also, this removes some extra closures that can be removed and corrects some more styles according to [scala-style-guide#pattern-matching](https://github.com/databricks/scala-style-guide#pattern-matching).

Lastly, `convertRow` is renamed to `convertObjectWithAttributes` and `convertValues` to `convertAttributes` just to be consistent with Spark (Parquet and Json data sources)


src/main/scala/com/databricks/spark/xml/parsers/StaxXmlParser.scala
```diff
@@ -211,13 +217,17 @@ private[xml] object StaxXmlParser {
...
     }
-    // The fields are sorted so `TreeMap` is used.
-    val convertedValuesMap = convertValues(valuesMap, schema)
-    val row = TreeMap((fields ++ convertedValuesMap).toSeq : _*).values.toSeq
+
+    // Here we merge both to a row.
+    val valuesMap = fieldsMap ++ attributesMap
+    valuesMap.foreach { case (f, v) =>
+      val nameToIndex = schema.map(_.name).zipWithIndex.toMap
+      nameToIndex.get(f).foreach(row.update(_, v))
```
lokm01: I think with an assignment to the row it'd be easier to spot what's happening here...

src/test/resources/simple-nested-objects.xml
```diff
@@ -0,0 +1,14 @@
...
+            <b>222</b>
+        </c>
+    </ROW>
+    <ROW>
+        <c>
+            <a>333</a>
+            <b>444</b>
+        </c>
+    </ROW>
+</ROWSET>
```
lokm01: Missing newline?

lokm01: Looks good!

src/main/scala/com/databricks/spark/xml/parsers/StaxXmlParser.scala
HyukjinKwon: I see. Thanks! I will also add a newline in the test data.

src/test/resources/simple-nested-objects.xml
HyukjinKwon: (Added)

lokm01:
Looks good :+1:

HyukjinKwon:
I am going to merge this after reviewing this multiple times by myself and @lokm01.

Issue #123, opened 2016-05-13
Change default values for valueTag and attributePrefix to avoid always require field escape for some apis.


HyukjinKwon:
If there is XML data as below:

``` xml
<row>
  <bbb id="1">2</bbb>
</row>
```

This library currently uses `#VALUE` for `valueTag` and `@` prefix for `attributePrefix` (they were just taken after a python library `xmltodict`)

So, the data above will produce a schema below:

```
root
 |-- bbb: struct (nullable = true)
 |    |-- #VALUE: long (nullable = true)
 |    |-- _id: long (nullable = true)
```

When it comes to Spark, it becomes problematic because it seems some APIs do not allow this field name. For example,

``` scala
val df = sqlContext.read
  .format("xml")
  .option("rowTag", "row")
  .load(path)

df.select("bbb.#VALUE").show()
```

this will work but 

``` scala
df.selectExpr("bbb.#VALUE").show()
```

does not work throwing an exception as below:

```
[1.11] failure: end of input

bbb.#VALUE
          ^
java.lang.RuntimeException: [1.11] failure: end of input

bbb.#VALUE
          ^
    at scala.sys.package$.error(package.scala:27)
    at org.apache.spark.sql.catalyst.SqlParser$.parseExpression(SqlParser.scala:49)
    at org.apache.spark.sql.DataFrame$$anonfun$selectExpr$1.apply(DataFrame.scala:785)
    at org.apache.spark.sql.DataFrame$$anonfun$selectExpr$1.apply(DataFrame.scala:784)
    at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:245)
    at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:245)
    at scala.collection.IndexedSeqOptimized$class.foreach(IndexedSeqOptimized.scala:33)
    at scala.collection.mutable.WrappedArray.foreach(WrappedArray.scala:35)
    at scala.collection.TraversableLike$class.map(TraversableLike.scala:245)
    at scala.collection.AbstractTraversable.map(Traversable.scala:104)
    at org.apache.spark.sql.DataFrame.selectExpr(DataFrame.scala:784)
```

It seems it would be safe to change this to `_VALUE` for `valueTag` and  `_` for `attributePrefix`. Spark uses `_c` prefix for unnamed columns.

HyukjinKwon:
[...]

PR #142, opened 2016-07-16
Change default values for valueTag and attributePrefix to avoid always require field escape for some apis.

HyukjinKwon:
#123 

This simply changes the default values in `valueTag` and `attributePrefix` due to some problems. Please refer the issue above for more details.

This PR also adds the test for default values for them. They were not being tested.


HyukjinKwon:
I am going to merge this.

PR #145, opened 2016-07-28
Add an alias `compression` for `codec` for consistency with Spark's internal data sources

HyukjinKwon:
Just for the consistency for internal Spark data sources (JSON and CSV), it matches the compression option name, `codec` to `compression`.


HyukjinKwon:
Merged.

PR #159, opened 2016-08-24
Change books to book in Hadoop InputFormat example

mattroberts297:
The example books.xml file does not contain a `books` element. It does
contain a `catalog` element and many `book` elements. Previously,
`records.count()` would return `0`. Now, it would return `12`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -479,7 +479,7 @@ sc.hadoopConfiguration.set(XmlInputFormat.START_TAG_KEY, "<book>")
 sc.hadoopConfiguration.set(XmlInputFormat.END_TAG_KEY, "</book>")
 sc.hadoopConfiguration.set(XmlInputFormat.ENCODING_KEY, "utf-8")
 
-val records = context.newAPIHadoopFile(
+val records = sc.newAPIHadoopFile(
   path,
   classOf[XmlInputFormat],
   classOf[LongWritable],
--- a/README.md
+++ b/README.md
@@ -475,8 +475,8 @@ which you may make direct use of as follows:
 import com.databricks.spark.xml.XmlInputFormat
 
 // This will detect the tags including attributes
-sc.hadoopConfiguration.set(XmlInputFormat.START_TAG_KEY, "<books>")
-sc.hadoopConfiguration.set(XmlInputFormat.END_TAG_KEY, "</books>")
+sc.hadoopConfiguration.set(XmlInputFormat.START_TAG_KEY, "<book>")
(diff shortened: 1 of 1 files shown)


HyukjinKwon:
Nice catch! Could you please check if there are similar instances in the documentation please?

mattroberts297:
Ah yes. You're correct there are a few other usages of `<books>` and `</books>` under the Features heading, but I think they should be changed to `<catalog>` and `</catalog>` respectively. I will update the PR.

mattroberts297:
Having looked at the diff of my update prior to commit. The usages of `<books>` and `</books>` under Features are currently correct because the examples are given inline and before the example books.xml file is linked to:

```diff
-* `rowTag`: The row tag of your xml files to treat as a row. For example, in this xml `<books> <book><book> ...</books>`, the appropriate value would be `book`. Default is `ROW`.
+* `rowTag`: The row tag of your xml files to treat as a row. For example, in this xml `<catalog> <book><book> ...</catalog>`, the appropriate value would be `book`. Default is `ROW`.
```

That said, they'd still be correct if I did update and consistent with books.xml. Completely up to you.

Also, in the API examples where the file is written out the root element is changed from `catalog` to `books`. I think it's nice to demonstrate that this can be done, so might be best to leave those as `books`. Unless your original intention was to keep the root element the same of course!

HyukjinKwon:
Yup, thanks. Let's just merge this one. Thanks! merging to master.

README.md
```diff
@@ -475,8 +475,8 @@ which you may make direct use of as follows:
 import com.databricks.spark.xml.XmlInputFormat
 
 // This will detect the tags including attributes
-sc.hadoopConfiguration.set(XmlInputFormat.START_TAG_KEY, "<books>")
-sc.hadoopConfiguration.set(XmlInputFormat.END_TAG_KEY, "</books>")
+sc.hadoopConfiguration.set(XmlInputFormat.START_TAG_KEY, "<book>")
+sc.hadoopConfiguration.set(XmlInputFormat.END_TAG_KEY, "</book>")
 sc.hadoopConfiguration.set(XmlInputFormat.ENCODING_KEY, "utf-8")
 
 val records = context.newAPIHadoopFile(
```
HyukjinKwon: Oh, wait, do you mind just to change `val records = context.newAPIHadoopFile(` to `val records = sc.newAPIHadoopFile(`? It'd be great if typos are fixed together in this example!

mattroberts297: Ah yes great spot. Will do...

HyukjinKwon:
Yeap, I am going to merge this. Thanks.

Issue #160, opened 2016-08-24
Nested elements with same name as parent fails


mattroberts297:
This XML:

```scala
<?xml version="1.0" encoding="UTF-8"?>
<body><p><p>Nested element</p><b>Important element</b></p></body>
```

When loaded using the XmlInputFormat (I think this underlies everything but may be wrong):

```scala
val records = sc.newAPIHadoopFile[LongWritable, Text, XmlInputFormat](path)
records.values.map(text => text.toString).foreach(log.info)
```

Logs this:

```bash
INFO Job$: <p><p>Nested element</p>
```

Instead of this:

```bash
<p><p>Nested element</p><b>Important element</b></p>
```

My workaround, use an xslt to rename the nested `<p>`:

```scala
<?xml version="1.0" encoding="UTF-8"?>
<body><p><a>Nested element</a><b>Important element</b></p>
</body>
```

I think the solution is to make `XmlRecordReader` track the depth (incrementing on start tags and decrementing on end tags) and only stop reading when the end tag is encountered and the depth is 0. Not entirely trivial and I think the bug exists in Mahout's Java implementation of [`XmlRecordReader`](https://github.com/apache/mahout/blob/ad84344e4055b1e6adff5779339a33fa29e1265d/examples/src/main/java/org/apache/mahout/classifier/bayes/XmlInputFormat.java) too. If you agree it's a bug I'll try and raise a PR for this.

On an aside, with the xslt'd version of my file the library works really well. Thanks!

HyukjinKwon:
Oh, yea. Thanks for detailed explanation. Yes, this is a bug and we should fix. Actually, this is a duplicate of https://github.com/databricks/spark-xml/issues/148 :)

Please go ahead for a PR. 

BTW, please make sure your commits are linked to you. I just merged your PR but realised that the commits are not linked to your.

HyukjinKwon:
Let me mark this as a duplicate. Let's close both if this one is solved (as this one has a good example and explanation)

mattroberts297:
> BTW, please make sure your commits are linked to you. I just merged your PR but realised that the commits are not linked to your github account.

Thanks for the heads up (bad config on my part).

HyukjinKwon:
This is fixed in #161

PR #161, opened 2016-08-25
Fix nested element with name of parent bug

### LINKED ISSUE #148: Dataframe is null when parent and child tag names are same
Following valid XML returns a null Dataframe:
&lt;Parent&gt;
  &lt;Child&gt;&lt;/Child&gt;
  &lt;Parent&gt;&lt;/Parent&gt;
&lt;/Parent&gt;

when row-tag is given as &lt;Parent&gt;

mattroberts297:
This should fix #148 and #160. It's not very elegant (or the easiest of reads for that matter), but I didn't want to change the performance characteristics of the code. The tests pass locally, so fingers crossed they pass on travis.

### DIFF
--- a/src/main/scala/com/databricks/spark/xml/XmlInputFormat.scala
+++ b/src/main/scala/com/databricks/spark/xml/XmlInputFormat.scala
@@ -168,26 +168,20 @@ private[xml] class XmlRecordReader extends RecordReader[LongWritable, Text] {
     }
   }
 
-  // TODO: Write a readUntilStartElement and readUntilEndElement
-
   private def readUntilStartElement(): Boolean = {
     @annotation.tailrec
     def loop(i: Int): Boolean = {
       val b = in.read()
-      // Check for EOF.
       if (b == -1 || (i == 0 && filePosition.getPos > end)) {
         false
       } else {
-        // Check for matching element byte.
         if (b == startTag(i)) {
-          // Check for end of match.
           if (i >= startTag.length - 1) {
             true
           } else {
             loop(i + 1)
           }
         } else {
-          // Check for attributes.
           if (i == (startTag.length - angleBracket.length) && checkAttributes(b)) {
             true
           } else {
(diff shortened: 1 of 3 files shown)


HyukjinKwon:
Thanks! @mattroberts297 I will run some small benchmarks with this and will review too on this weekend. Thanks for working on this.

HyukjinKwon:
Generally looks good and neat to me. I am running small benckmarks with wiki although it looks it would not affect the performance. After this, I will take a closer look.

HyukjinKwon:
I ran the codes below and calculate the average time taken after running 5 times each.

```scala
test("simple benchmarks") {
  def time[A](f: => A) = {
    val s = System.nanoTime
    val ret = f
    println("time: "+(System.nanoTime-s)/1e6+"ms")
    ret
  }

  val config = new Configuration(sqlContext.sparkContext.hadoopConfiguration)
  config.set(XmlInputFormat.START_TAG_KEY, "<page>")
  config.set(XmlInputFormat.END_TAG_KEY, "</page>")
  config.set(XmlInputFormat.ENCODING_KEY, "utf-8")
  val records = sqlContext.sparkContext.newAPIHadoopFile(
    wikiFilePath,
    classOf[XmlInputFormat],
    classOf[LongWritable],
    classOf[Text],
    config)


  time {
    records.count()
  }
}
```

**Before** - 136763.142486 ms

**After** - 136915.789208 ms

It seems the difference is ignorable. I look though some blogs and references for the performance between tail recursion vs while loop and it seems arguable which one is faster. Therefore, it seems generally okay.  However, it seems it is better to avoid recursion as mentioned here, https://github.com/databricks/scala-style-guide#recursion .

src/main/scala/com/databricks/spark/xml/XmlInputFormat.scala
```diff
@@ -168,46 +168,64 @@ private[xml] class XmlRecordReader extends RecordReader[LongWritable, Text] {
...
-        buffer.write(b)
-      }
-      if (b == mat(i)) {
-        i += 1
-        if (i >= mat.length) {
-          currentStartTag = startTag
-          return true
+  private def readUntilStartElement(): Boolean = {
+    @annotation.tailrec
+    def loop(i: Int): Boolean = {
```
HyukjinKwon: So, personally I like the changes in general but I am just wondering if we can avoid recursion but use just while loop.

mattroberts297: I could avoid tail recursion and use a while loop instead, but missing cases could only be caught at runtime (with tests) as opposed to compile time (with the compiler). I would also need to use mutable variables and return statements both of which break referential transparency making the code harder to reason about and refactor. In short, I would recommend against it. That said, I appreciate this is a DataBricks repository and their style guide explicitly recommends against tail recursion:

> Avoid using recursion, unless the problem can be naturally framed recursively (e.g. graph traversal, tree traversal).

With the above in mind, would you still like me to rewrite using a while loop?

Regarding comments, I am happy to put those in whether using tail recursion or a while loop. I know some people don't like them, but the code is quite low level, so I agree they will help. Do you prefer above or to the right (subject to line length)?

HyukjinKwon: If this were my private repo, I'd be happy to accept this PR as it is but I am worried because this one is now under Databricks (and I am actually not a Databricks guy but just an individual). So, please do me a favour to stay in a safe-side.

Also, actually. I am about to release a new one. So, let's use loop one for now and then include that in the release. I would like to include this change. After that, maybe we can make another PR to use recursive one with cc'ing other guys to discuss further.

HyukjinKwon: As soon as you fix up, I will run another benchmarks again to verify performance regression quickly.

mattroberts297: Sounds good and understood I will try and fix up this afternoon. :+1:

HyukjinKwon:
Thanks for bearing with me @mattroberts297. I will post the results of benchmarks here tomorrow.

LGTM but I will take another look to make sure (because we should be careful of such changes as you already know).

HyukjinKwon:
It seems something gone wrong with my test environment. So, I ended up with testing all the versions again as below (run 5 times and averages):

- Original one : 144655.5546866ms

- Recursive one : 146596.1500502ms

- Normal loop one : 147619.160973ms

Again, it seems ignorable and it looks the best effort. I am merging this.

HyukjinKwon:
Thanks @mattroberts297 again.

AshwaniGaur:
This bug is still not fixed with large partition. If child tag (having same name as parent tag) is found in another partition, it is treated as parent tag.

PR #166, opened 2016-08-31
Add missed other default case when parsing/inferring XML documents

HyukjinKwon:
This PR adds the support for skipping multiple white spaces around a comment.

This should have been added but missed. As `XMLStreamConstants.COMMENT` is always skipped [here](https://github.com/databricks/spark-xml/blob/master/src/main/scala/com/databricks/spark/xml/parsers/StaxXmlParser.scala#L51-L52) and [here](https://github.com/databricks/spark-xml/blob/master/src/main/scala/com/databricks/spark/xml/util/InferSchema.scala#L85-L86) but it seems it is possible to have the `COMMENT` is between white spaces.

In this case, `factory.setProperty(XMLInputFactory.IS_COALESCING, true)` does not coalesce both white spaces.

In more details,

```xml
<a>
  <!-- comment -->
  <b>...</b>
</a>
```

in this case, `<!--comment -->` is surrounded with whitespaces. 

This produces the events as blow:

```bash
XMLStreamConstants.CHARACTERS     # whitespace
XMLStreamConstants.COMMENT        # comment
XMLStreamConstants.CHARACTERS     # whitespace
XMLStreamConstants.START_ELEMENT  # <b>
```

Current codes always filter `XmlEvent.COMMENT` so it ends up with 

```bash
XMLStreamConstants.CHARACTERS     # whitespace
XMLStreamConstants.CHARACTERS     # whitespace
XMLStreamConstants.START_ELEMENT  # <b>
```

which does not happen in normal cases because we are coalescing multiple `XMLStreamConstants.CHARACTERS` into single one as below:

```bash
XMLStreamConstants.CHARACTERS     # whitespace
XMLStreamConstants.START_ELEMENT  # <b>
```

### DIFF
--- a/src/main/scala/com/databricks/spark/xml/parsers/StaxXmlParser.scala
+++ b/src/main/scala/com/databricks/spark/xml/parsers/StaxXmlParser.scala
@@ -112,6 +112,7 @@ private[xml] object StaxXmlParser {
           case _: EndElement if data.isEmpty => null
           case _: EndElement if options.treatEmptyValuesAsNulls => null
           case _: EndElement => data
+          case _ => convertField(parser, dataType, options)
         }
 
       case (c: Characters, ArrayType(st, _)) =>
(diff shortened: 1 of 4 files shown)


HyukjinKwon:
I am going to merge this.

PR #170, opened 2016-09-04
Do not allow empty strings for attributePrefix, valueTag and rowTag

HyukjinKwon:
This PR adds requirements for `attributePrefix`, `valueTag` and `rowTag`.

Since those options are used to differentiate attributes, `attributePrefix` should not be just a empty string

`valueTag` is used for a column name so it should not be empty string.

Lastly, we should set the explicit range for each row via `rowTag`. Therefore, this should not be just empty string too.


HyukjinKwon:
I am going to merge this.

Issue #171, opened 2016-09-04
XmlInputFormat produces duplicate rows in large files


lokm01:
There's a bug in the XmlInputFormat which causes duplicate rows to be produced when the input file spans over more input splits (for plain text and splittable compression format) in cases, where there is no whitespace between closing and opening row tags.

How to reproduce:
1) Use large XML file (e.g the protein file from http://www.cs.washington.edu/research/xmldatasets/www/repository.html#pir )
2) Modify to remove whitespace between the closing and opening row tags (e.g </ProteinEntry><ProteinEntry id="CCCZ" > )
3) See that the rows are duplicated

This is due to the fact that the condition which checks the end of the file position isn't reached as the check of the start tag returns true before getting to that point.

HyukjinKwon:
I will try to reproduce this soon. Thanks for reporting this.

lokm01:
This seems to be fixed by #161 
There is however another bug which appeared after this PR on the same data when using compression codecs.

`
org.apache.spark.SparkException: Job aborted due to stage failure: Task 1 in stage 123.0 failed 1 times, most recent failure: Lost task 1.0 in stage 123.0 (TID 131, localhost): java.lang.ArrayIndexOutOfBoundsException: 65536
[info] 	at org.apache.hadoop.io.compress.bzip2.CBZip2InputStream.hbCreateDecodeTables(CBZip2InputStream.java:666)
[info] 	at org.apache.hadoop.io.compress.bzip2.CBZip2InputStream.createHuffmanDecodingTables(CBZip2InputStream.java:793)
[info] 	at org.apache.hadoop.io.compress.bzip2.CBZip2InputStream.recvDecodingTables(CBZip2InputStream.java:765)
[info] 	at org.apache.hadoop.io.compress.bzip2.CBZip2InputStream.getAndMoveToFrontDecode(CBZip2InputStream.java:801)
[info] 	at org.apache.hadoop.io.compress.bzip2.CBZip2InputStream.initBlock(CBZip2InputStream.java:504)
[info] 	at org.apache.hadoop.io.compress.bzip2.CBZip2InputStream.changeStateToProcessABlock(CBZip2InputStream.java:333)
[info] 	at org.apache.hadoop.io.compress.bzip2.CBZip2InputStream.read(CBZip2InputStream.java:399)
[info] 	at org.apache.hadoop.io.compress.BZip2Codec$BZip2CompressionInputStream.read(BZip2Codec.java:483)
[...]

PR #174, opened 2016-09-04
Fixing duplicate rows issue in XmlInputFormat #171

lokm01:
Fix for issue #171

### DIFF
--- a/src/main/scala/com/databricks/spark/xml/XmlInputFormat.scala
+++ b/src/main/scala/com/databricks/spark/xml/XmlInputFormat.scala
@@ -65,7 +65,8 @@ private[xml] class XmlRecordReader extends RecordReader[LongWritable, Text] {
   private var in: InputStream = _
   private var filePosition: Seekable = _
   private var decompressor: Decompressor = _
-
+  private var compressRatio: Double = _
+  
   private val buffer: DataOutputBuffer = new DataOutputBuffer
 
   override def initialize(split: InputSplit, context: TaskAttemptContext): Unit = {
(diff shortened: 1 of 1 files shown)


lokm01:
@HyukjinKwon

HyukjinKwon:
Hi @lokm01 , I didn't take a look so closely but I resembled [LineRecordReader.java#L94-L134](https://github.com/apache/hadoop/blob/branch-2.2.0/hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-core/src/main/java/org/apache/hadoop/mapred/LineRecordReader.java#L94-L134) in Hadoop 2.2.0. However, in a recent version, I noticed it was changed to [LineRecordReader.java#L97-L146](https://github.com/apache/hadoop/blob/branch-3.0.0-alpha1/hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-core/src/main/java/org/apache/hadoop/mapred/LineRecordReader.java#L97-L146) in Hadoop 3.0.0. And assuming from the related commit, https://github.com/apache/hadoop/commit/9227dfc25f373a99cb66ad7d6bacef8dcf336f77, it seems the issue is similar?

Could you maybe check if the problem is the same and resemble the way if so? I would like to track down the changes there rather than duplicating the efforts.

HyukjinKwon:
BTW, maybe it seems you have something wrong with updating your PR. I will leave some notes for this just in case. Maybe you could do as below:

```
git remote add upstream https://github.com/databricks/spark-xml.git
git fetch upstream
git checkout issue-171
git merge upstream/master
git push origin issue-171
```

There are several ways to update your PR such as rebasing or squashing. I found this reference is good, https://github.com/druid-io/druid/blob/master/CONTRIBUTING.md

HyukjinKwon:
ping @lokm01 If you don't have some time, I will do this maybe within next few weeks.

lokm01:
Will have a look during the week.

lokm01:
I am closing this PR. #161 seems to fix this particular issue, though I found another problem with compressed files which appeared with #161 need to investigate this further.

PR #150, opened 2016-08-01
Upgrade spark 2.0 and prepare 0.4.0-SNAPSHOT

HyukjinKwon:
This PR migrates spark-xml to Spark 2.0.

This PR,
 - deprecates `saveAsXmlFile` and promote the usage of `write()`.
 - deprecates `xmlFile` and promote the usage of `read()`.
 - drops 1.x compatibility from 0.4.0.
 - makes not supporting `UserDefinedType` as it became private.

PR #107, opened 2016-04-04
Support for PERMISSIVE/DROPMALFORMED mode and corrupt record option.

HyukjinKwon:
https://github.com/databricks/spark-xml/issues/105

Currently, this library does not support `PERMISSIVE` parse mode. Similar with JSON data source, this also can be done in the same way with `_corrupt_record`.

This PR adds the support for `PERMISSIVE` mode and make this behaviour consistent with the other data sources supporting parse modes (JSON and CSV data sources.)

Also, this PR adds the support for `_corrupt_record`.

This PR is similar with https://github.com/apache/spark/pull/11756 and https://github.com/apache/spark/pull/11881.

PR #176, opened 2016-09-10
Changes for 0.4.0

HyukjinKwon:
This PR prepares the release for 0.4.0.

This will include the changes below:
  - Support for PERMISSIVE/DROPMALFORMED mode and corrupt record option. #107
  - Change default values for valueTag and attributePrefix to avoid always require field escape for some apis #142 
  - deprecates saveAsXmlFile and promote the usage of write(). #150
  - deprecates xmlFile and promote the usage of read(). #150
  - drops 1.x compatibility from 0.4.0. #150
  - makes not supporting UserDefinedType as it became private. #150

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,23 +10,28 @@ The structure and test tools are mostly copied from [CSV Data Source for Spark](
 
 ## Requirements
 
-This library requires Spark 1.3+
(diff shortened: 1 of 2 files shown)


HyukjinKwon:
I am going to merge this.

Issue #185, opened 2016-10-05
Try parser XML


luissantanayh:
**I have this schema:**

root
 |-- _Id: string (nullable = true)
 |-- _versao: double (nullable = true)
 |-- autXML: struct (nullable = true)
 |    |-- CNPJ: long (nullable = true)
 |-- dest: struct (nullable = true)
 |    |-- CNPJ: long (nullable = true)
 |    |-- IE: long (nullable = true)
 |    |-- enderDest: struct (nullable = true)
 |    |    |-- CEP: long (nullable = true)
 |    |    |-- UF: string (nullable = true)
 |    |    |-- cMun: long (nullable = true)
 |    |    |-- cPais: long (nullable = true)
 |    |    |-- fone: long (nullable = true)
 |    |    |-- nro: string (nullable = true)
 |    |    |-- xBairro: string (nullable = true)
 |    |    |-- xLgr: string (nullable = true)
 |    |    |-- xMun: string (nullable = true)
 |    |    |-- xPais: string (nullable = true)
 |    |-- indIEDest: long (nullable = true)
 |    |-- xNome: string (nullable = true)
 |-- det: struct (nullable = true)
 |    |-- _nItem: long (nullable = true)
 |    |-- imposto: struct (nullable = true)
 |    |    |-- COFINS: struct (nullable = true)
 |    |    |    |-- COFINSAliq: struct (nullable = true)
 |    |    |    |    |-- CST: long (nullable = true)
 |    |    |    |    |-- pCOFINS: double (nullable = true)
 |    |    |    |    |-- vBC: double (nullable = true)
 |    |    |    |    |-- vCOFINS: double (nullable = true)
 |    |    |-- ICMS: struct (nullable = true)
 |    |    |    |-- ICMS40: struct (nullable = true)
 |    |    |    |    |-- CST: long (nullable = true)
 |    |    |    |    |-- orig: long (nullable = true)
 |    |    |-- PIS: struct (nullable = true)
 |    |    |    |-- PISAliq: struct (nullable = true)
 |    |    |    |    |-- CST: long (nullable = true)
 |    |    |    |    |-- pPIS: double (nullable = true)
 |    |    |    |    |-- vBC: double (nullable = true)
 |    |    |    |    |-- vPIS: double (nullable = true)
 |    |-- prod: struct (nullable = true)
 |    |    |-- CFOP: long (nullable = true)
 |    |    |-- NCM: long (nullable = true)
[...]

PR #186, opened 2016-10-06
Do not sienlty produce null but skip the children XML elements by the pushed down projection

HyukjinKwon:
https://github.com/databricks/spark-xml/issues/185

This PR proposes to handle pushed down projection correctly via skipping the children XML elements.

`PrunedScan` was introduced in XML before and therefore we should be able to skip the children elements. Currently, it sometimes for parsing and skipping the elements of struct types when processing pushed down projection.


HyukjinKwon:
I am going to merge this.

PR #220, opened 2016-12-16
Write values properly as given to `valueTag` when it conflicts to `attributePrefix`

HyukjinKwon:
If `valueTag` and `attributePrefix` are conflicted (meaning `valueTag` starting with `attributePrefix`, it writes out the data in an incorrect format.

For example, running the codes

```scala
val path = "path-to-file"
books.write
  .format("xml")
  .option("valueTag", "#VALUE")
  .option("attributePrefix", "#")
  .save(path) 
```

ends up with 

```xml
...
<Values VALUE="0.56"></Values>
...
```

After this PR, 

```xml
...
<Values>0.56</Values>
...
```


HyukjinKwon:
I am going to merge this after few days.

HyukjinKwon:
It should be backported into branch-0.3.

PR #228, opened 2017-01-22
is there any way to refer XSD as a schema reference to process xml messages ? It should handle complex type like arraytype and structType based on minoccurs,maxoccurs and unbounded.

skm235:


HyukjinKwon:
Let me close this. It looks mistakenly opened.

PR #232, opened 2017-01-27
Added documentation entry to warn against self closing row tags

trheming:


HyukjinKwon:
Cool. It is always good to document the limitation. Thanks. Could you maybe copy and paste it to `rowTag` in the writing options below?

trheming:
Not positive, but my understanding is that the writer will never be able to encounter this condition when saving out a dataframe? Things would default to `<rowTag> </rowTag>`. Is this correct?

HyukjinKwon:
Oh, yes. I was thinking when someone might want to write out with self-closing tags by kind of an option. Yea, but it sounds less common. LGTM. Let me merge this tomorrow (due to network problem).

HyukjinKwon:
Thank you @trheming, merging to master.

Issue #236, opened 2017-02-09
The fields with extra whitespaces have correctly passed through the XML validation against their correct types (xsd:double, xsd:dateTime, etc) but fail during parsing in spark-xml


kkaduk:
Elements in xml document for example  &lt; tagA &gt; 12.33  &lt; /tagA &gt; (trailing or leading spaces ) are not correctly converted with exception:
java.lang.NumberFormatException: For input string: "  12.33   "
        at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
        at java.lang.Long.parseLong(Long.java:441)

Also kills the spark executor and fails the entire Spark job.

HyukjinKwon:
Thanks for reporting this.

HyukjinKwon:
Fixed in #237

PR #237, opened 2017-02-11
Add an option `ignoreSurroundingSpaces` to allow to trim spaces between values

HyukjinKwon:
https://github.com/databricks/spark-xml/issues/236

It seems it is painful to treat these, especially, we provide a schema to XML file but it ended up with having some spaces. It ends up with an error such as ...

```
java.lang.NumberFormatException: For input string: "1 "
        at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
        at java.lang.Long.parseLong(Long.java:441)
```

It seems supporting to trim seems reasonable according to https://www.w3.org/TR/xmlschema-2/#rf-whiteSpace

### DIFF
--- a/src/main/scala/com/databricks/spark/xml/util/TypeCast.scala
+++ b/src/main/scala/com/databricks/spark/xml/util/TypeCast.scala
@@ -75,7 +75,12 @@ object TypeCast {
       datum: String,
       dataType: DataType,
       options: XmlOptions): Any = {
-    val value = if (options.ignoreSurroundingSpaces) datum.trim() else datum
+    val value = if (datum != null && options.ignoreSurroundingSpaces) {
+      datum.trim()
+    } else {
+      datum
+    }
+
     dataType match {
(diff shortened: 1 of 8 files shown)


HyukjinKwon:
(Thank you @kkaduk for providing some references and use cases).

src/main/scala/com/databricks/spark/xml/util/TypeCast.scala
```diff
@@ -70,6 +70,35 @@ object TypeCast {
     }
   }
 
+  // TODO: This function unnecessarily does type dispatch. Should merge it with `castTo`.
+  private[xml] def convertTo(
```
HyukjinKwon: It is from `StaxXmlParser.convertTo`.

HyukjinKwon:
Let me leave this open for few days more before merging this.

HyukjinKwon:
Merged into master and branch-0.4. Let me open a backport for branch-0.3.

PR #247, opened 2017-03-22
Fix for NPE and support for attributes at root level

fmcgough:
- Fixes a NullPointerException when writing null attribute values at row level
- Adds support for adding attributes to the root element. The main use case for this is adding `xmlns` attributes; this is implemented by means of adding options that start with the attribute prefix.
- Extracts the functionality for creating an RDD[String] from a dataframe into its own object. My motivation for this change was a data import format that requires a `<Summary>` element at the start of the XML document, but it also allows for arbitrary manipulation of the data as strings. I haven't documented this in the readme as this is a fairly advanced use case - happy to add a bit in if you think it's appropriate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,7 @@ When writing files the API accepts several options:
 * `attributePrefix`: The prefix for attributes so that we can differentiating attributes and elements. This will be the prefix for field names. Default is `_`.
 * `valueTag`: The tag used for the value when there are attributes in the element having no child. Default is `_VALUE`.
 * `compression`: compression codec to use when saving to file. Should be the fully qualified name of a class implementing `org.apache.hadoop.io.compress.CompressionCodec` or one of case-insensitive shorten names (`bzip2`, `gzip`, `lz4`, and `snappy`). Defaults to no compression when a codec is not specified.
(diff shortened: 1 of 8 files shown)


HyukjinKwon:
Thank you for your proposal @fmcgough. Let me take a look soon.

HyukjinKwon:
> - Fixes a NullPointerException when writing null attribute values at row level

Yes, we should definitely fix this.

> - Adds support for adding attributes to the root element. The main use case for this is adding `xmlns` attributes; this is implemented by means of adding options that start with the attribute prefix.

This case seems allowing the option `"_xmlns" -> "http://example.com/books"`. I am worried of allowing arbitrary keys. I think we should find another way. For example, allowing the raw string maybe?

> - Extracts the functionality for creating an RDD[String] from a dataframe into its own object. My motivation for this change was a data import format that requires a `<Summary>` element at the start of the XML document, but it also allows for arbitrary manipulation of the data as strings. I haven't documented this in the readme as this is a fairly advanced use case - happy to add a bit in if you think it's appropriate.

I am not positive of this change. I think we should avoid exposing other ways to use this library other than `spark.write.format("xml")`. Other varints should be covered via options if possible or handled in user's application to reduce the maintain cost.

This functionality sounds like `to_json` function in Spark. How about raising a JIRA in Spark or asking this into mailing list?

In general, do you mind if I ask split this PRs? For the first case, we should definitely fix. For second case, I think we should discuss about how to support this. For the last one, I think we should better raise an issue or ask a question to dev/user mailing list within Spark. I would like to avoid adding many ways to use this library.

PR #282, opened 2018-02-09
Output valid streaming XML with woodstox

pwoody:
Hadoop added a woodstox dependency in https://issues.apache.org/jira/browse/HADOOP-14501. When using versions of Hadoop with that dependency, WstxOutputFactory will get loaded for XMLOutputFactory.newInstance. The resulting streamwriter will ensure that a client does not write two root elements, which causes the current code to error out.

This PR makes it such that the writer function will start the root tag before writing rows.

### DIFF
--- a/src/main/scala/com/databricks/spark/xml/util/XmlFile.scala
+++ b/src/main/scala/com/databricks/spark/xml/util/XmlFile.scala
@@ -101,6 +101,10 @@ private[xml] object XmlFile {
 
         override def next: String = {
           if (iter.nonEmpty) {
+            if (firstRow) {
+              indentingXmlWriter.writeStartElement(options.rootTag)
+              firstRow = false
+            }
             val xml = {
               StaxXmlGenerator(
                 rowSchema,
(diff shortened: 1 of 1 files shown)


pwoody:
@HyukjinKwon let me know what you think. Thanks!

HyukjinKwon:
Wow, thanks for fixing it @pwoody. Seems fine. Let me merge this one after another double check within few days.

src/main/scala/com/databricks/spark/xml/util/XmlFile.scala
```diff
@@ -101,6 +101,10 @@ private[xml] object XmlFile {
```
HyukjinKwon: @pwoody, mind if I ask to remove `val startElement = s"<${options.rootTag}>"` and `val endElement = s"</${options.rootTag}>"` above? Seems unused.

pwoody: Yep, done.

HyukjinKwon:
Merged to master and branch-0.4!

pwoody:
Thanks @HyukjinKwon - what is the release cadence for a 0.4.x release?

HyukjinKwon:
Will make a release soon within next week (after sorting out the current open PRs). Does that work for you too?

pwoody:
Yep, no rush!

HyukjinKwon:
Hey, @pwoody. I'm really sorry that the release has been postponed. 0.5.0 is released now. Thank you for your work .. !

PR #286, opened 2018-03-09
Fix parsing xml with unicode characters in row tag name

neherim:
Currently, only supported english letters in «rowTag» option.
But non-English letters and unicode characters are perfectly legal in XML element names
https://www.w3.org/TR/xml/#charsets

This PR adds the support for non-english letters and unicode characters in tag names passed as «rowTag» param in options

### DIFF
--- a/src/main/scala/com/databricks/spark/xml/XmlInputFormat.scala
+++ b/src/main/scala/com/databricks/spark/xml/XmlInputFormat.scala
@@ -164,7 +164,7 @@ private[xml] class XmlRecordReader extends RecordReader[LongWritable, Text] {
         // End of file or end of split.
         return false
       } else {
-        if (b == startTag(i)) {
+        if (b.toByte == startTag(i)) {
           if (i >= startTag.length - 1) {
             // Found start tag.
             return true
@@ -192,12 +192,13 @@ private[xml] class XmlRecordReader extends RecordReader[LongWritable, Text] {
     var ei = 0
     var depth = 0
     while (true) {
-      val b = in.read()
-      if (b == -1) {
+      val rb = in.read()
+      if (rb == -1) {
         // End of file (ignore end of split).
         return false
       } else {
-        buffer.write(b)
+        buffer.write(rb)
+        val b = rb.toByte
         if (b == startTag(si) && b == endTag(ei)) {
           // In start tag or end tag.
           si += 1
--- a/src/test/resources/books-unicode-in-tag-name.xml
+++ b/src/test/resources/books-unicode-in-tag-name.xml
@@ -1,0 +1,24 @@
+<?xml version="1.0"?>
+<κατάλογος>
+    <書 ид="bk101">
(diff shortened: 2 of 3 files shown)


src/main/scala/com/databricks/spark/xml/XmlInputFormat.scala
```diff
@@ -164,7 +164,7 @@ private[xml] class XmlRecordReader extends RecordReader[LongWritable, Text] {
         // End of file or end of split.
         return false
       } else {
-        if (b == startTag(i)) {
+        if (b.toByte == startTag(i)) {
```
HyukjinKwon: Sorry, @neherim, mind elaborating how it fixes the issue?

neherim:
In `XmlRecordReader.readUntilStartElement()`, we sequentially read bytes from the xml and compare them with `startTag` array.

For example we have `rowTag = 書`. This character in the UTF-8 encoding is represented as 3 bytes: `-26, -101, -72`.
So `startTag` array will be:
```Scala
// startTag: Array[Byte]
startTag[0] = 60   // ‘<’
startTag[1] = -26  // 書
startTag[2] = -101 // 書
startTag[3] = -72  // 書
startTag[4] = 62   // ‘>’
```

If we sequentially read this tag `<書>` from xml file with `InputStream.read()`, then we get following array:
```Scala
// tagFromXmlFile: Array[Int]
tagFromXmlFile[0] = 60  // ‘<‘
tagFromXmlFile[1] = 230 // 書
tagFromXmlFile[2] = 155 // 書
tagFromXmlFile[3] = 184 // 書
tagFromXmlFile[4] = 62  // ‘>’
```

The problem is that the function `read()` returns result as an **Int** in the range 0 to 255 ([javadocs](https://docs.oracle.com/javase/8/docs/api/java/io/InputStream.html#read--)).
But startTag array contains Bytes, and Byte value range is -128 to 127

So, when we try to compare the bytes from `startTag` and the values returned by `InputStream.read()`, the following happens:
```Scala
startTag[0] == tagFromXmlFile[0]  // true 60 == 60
startTag[1] == tagFromXmlFile[1]  // false -26 != 230
startTag[2] == tagFromXmlFile[2]  // false -101 != 155
startTag[3] == tagFromXmlFile[3]  // false -72 != 184
startTag[4] == tagFromXmlFile[4]  // true 62 == 62
```

And thats way we need explicitly convert `InputStream.read()` result to Byte, before comparing with `startTag`

src/main/scala/com/databricks/spark/xml/XmlInputFormat.scala
neherim: Of course, described fix in comments.

HyukjinKwon:
Will merge this in few more days.

HyukjinKwon:
Merged to master and branch-0.4.

Thanks for taking a look for this @neherim.

Issue #235, opened 2017-02-03
Failed type cast with provided schema causes all fields to be null


harrisonmebane:
When providing a schema, if the schema specifies a type for a field, and that field has a value that cannot be cast to that type, all fields in the resulting DataFrame are null.  The example below is minimal, but I observed the same error with an XML of thousands of lines; a single incorrect type caused the entire DataFrame to be populated with nulls.  This is `spark-xml` version `0.3.5`, on Spark 1.6.

Example:

XML:
```xml
<TopLevel>
  <StringField>abc123</StringField>
  <IntField>xyz456</IntField>
</TopLevel>
```

Schema:
```scala
val simpleSchema = 
StructType(
  Array(
    StructField("StringField",StringType,true),
    StructField("IntField",IntegerType,true)
  )
)
```

Result:
```scala
val simpleTest = sqlContext.read.format("com.databricks.spark.xml").option("rowTag", "TopLevel").schema(simpleSchema).load("test.xml")
simpleTest.show
```
```
+-----------+--------+
|StringField|IntField|
+-----------+--------+
|       null|    null|
+-----------+--------+
```

If you can think of any workarounds in the meantime, that would also be helpful.  The only thing I can think of is to change all types in the static schema to strings, but then it is impossible to do arithmetic in the resulting DataFrame.

HyukjinKwon:
Thanks for detailed explanation. Will take a look soon in more details.

Volodymyr128:
Was it fixed? I have the same issue for Spark 2.4.4

PR #294, opened 2018-04-05
Increase nulling granularity to preserve valid data

Woffendm:
Encountering field parsing errors with a user-supplied schema no longer nullifies all columns in a row; instead, it only nulls out the field that failed to parse (Addresses issue #235 open since Feb 2017)

### DIFF
--- a/src/main/scala/com/databricks/spark/xml/util/TypeCast.scala
+++ b/src/main/scala/com/databricks/spark/xml/util/TypeCast.scala
@@ -80,22 +80,33 @@ object TypeCast {
     } else {
       datum
     }
-
-    dataType match {
-      case NullType => castTo(value, StringType, options)
-      case LongType => signSafeToLong(value, options)
-      case DoubleType => signSafeToDouble(value, options)
-      case BooleanType => castTo(value, BooleanType, options)
-      case StringType => castTo(value, StringType, options)
-      case DateType => castTo(value, DateType, options)
-      case TimestampType => castTo(value, TimestampType, options)
-      case FloatType => signSafeToFloat(value, options)
-      case ByteType => castTo(value, ByteType, options)
-      case ShortType => castTo(value, ShortType, options)
-      case IntegerType => signSafeToInt(value, options)
-      case dt: DecimalType => castTo(value, dt, options)
-      case _ =>
-        sys.error(s"Failed to parse a value for data type $dataType.")
+    try {
+      dataType match {
+        case NullType => castTo(value, StringType, options)
+        case LongType => signSafeToLong(value, options)
+        case DoubleType => signSafeToDouble(value, options)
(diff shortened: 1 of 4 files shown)


src/main/scala/com/databricks/spark/xml/util/TypeCast.scala
```diff
@@ -80,22 +80,33 @@ object TypeCast {
...
+        case ByteType => castTo(value, ByteType, options)
+        case ShortType => castTo(value, ShortType, options)
+        case IntegerType => signSafeToInt(value, options)
+        case dt: DecimalType => castTo(value, dt, options)
+        case _ =>
+          sys.error(s"Failed to parse a value for data type $dataType.")
+      }
+    }
+    catch {
+      case e: Exception => {
```
HyukjinKwon: We could do this:

```
case NonFatal(_) if options.permissive => null
```

Woffendm: Done.

Woffendm:
I implemented your suggestion.

Woffendm:
Any updates?

gatorsmile:
cc @HyukjinKwon

src/main/scala/com/databricks/spark/xml/util/TypeCast.scala
```diff
@@ -22,10 +22,11 @@ import java.util.Locale
 
 import scala.util.Try
 import scala.util.control.Exception._
-
 import org.apache.spark.sql.types._
 import com.databricks.spark.xml.XmlOptions
 
+import scala.util.control.NonFatal
```
HyukjinKwon: ditto for import style - https://github.com/databricks/scala-style-guide#imports

src/main/scala/com/databricks/spark/xml/util/TypeCast.scala
```diff
@@ -80,22 +81,26 @@ object TypeCast {
...
+        case ByteType => castTo(value, ByteType, options)
+        case ShortType => castTo(value, ShortType, options)
+        case IntegerType => signSafeToInt(value, options)
+        case dt: DecimalType => castTo(value, dt, options)
+        case _ =>
+          sys.error(s"Failed to parse a value for data type $dataType.")
+      }
+    }
+    catch {
+      case NonFatal(_) if options.permissive => null
```
HyukjinKwon: Hmmmmm.. IIRC Spark's JSON doesn't also allow the partial results. Also, wouldn't this fail to fill up the corrupt column?

Woffendm:
My team is no longer using the SparkXML library for our ETL process; however, I'll continue to work on these pull requests as time allows. It may take a while though since we have upcoming deadlines for key deliverables.

HyukjinKwon:
Fixed in #358

PR #298, opened 2018-04-16
Dataframe parsing

Woffendm:
Adds the ability to parse XML stored as a string column in an RDD of Rows, and return a dataset of structs.


src/main/scala/com/databricks/spark/xml/XmlReader.scala
```diff
@@ -19,10 +19,11 @@ import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.{DataFrame, SQLContext}
 import org.apache.spark.sql.types.StructType
 import com.databricks.spark.xml.util.XmlFile
+import org.apache.spark.sql.Row;
 
 /**
- * A collection of static functions for working with XML files in Spark SQL
- */
+  * A collection of static functions for working with XML files in Spark SQL
+  */
```
HyukjinKwon: Spark itself follows Java style comment. Should be better to keep it consistent with Apache Spark.

Woffendm:
My team is no longer using the SparkXML library for our ETL process; however, I'll continue to work on these pull requests as time allows. It may take a while though since we have upcoming deadlines for key deliverables.

PR #316, opened 2018-07-08
Add test case for self-closing tag inside row

gerbszt:
The [latest commit](https://github.com/databricks/spark-xml/commit/f3772b296d77a5a92ae4e714f3f8cb9940f1f0f9) to master branch introduced a small bug - rows with self-closing tags inside cannot be parsed anymore.

XmlInputFormat stops reading the row once it encounters '/>' sequence. It was meant to be one of the possible endings of row, but it may happen inside it as well.

This PR doesn't solve the problem, but makes it easy to spot by providing a simple test case.
Previous master's HEAD (prior the commit in question) passes the test, so maybe the best approach would be to revert f3772 commit, apply this PR and merge f3772 back once it passes updated test suite.

### DIFF
--- a/src/test/resources/self-closing-tag.xml
+++ b/src/test/resources/self-closing-tag.xml
@@ -1,0 +1,6 @@
+<ROWSET>
+    <ROW>
+        <non-empty-tag>1</non-empty-tag>
+        <self-closing-tag/>
+    </ROW>
+</ROWSET>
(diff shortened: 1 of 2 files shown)


HyukjinKwon:
Let me revert that one and make a release within few weeks as it's really been a while since the last release.

BioQwer:
@HyukjinKwon I'm sorry for the delay. 
Wait until Monday please. After monday you can revert.
Is ok?

srowen:
This should be fixed by https://github.com/databricks/spark-xml/pull/352

PR #321, opened 2018-08-01
Fix case for self closing tag inside row

BioQwer:
fix #316


BioQwer:
@HyukjinKwon I fixed this bug
merge this plz

BioQwer:
@srowen can your review this bug fix?

src/main/scala/com/databricks/spark/xml/XmlInputFormat.scala
```diff
@@ -25,14 +25,17 @@ import org.apache.hadoop.io.{DataOutputBuffer, LongWritable, Text}
 import org.apache.hadoop.mapreduce.{InputSplit, RecordReader, TaskAttemptContext}
 import org.apache.hadoop.mapreduce.lib.input.{FileSplit, TextInputFormat}
 
+import scala.collection.mutable.ArrayBuffer
+
 /**
- * Reads records that are delimited by a specific start/end tag.
- */
+  * Reads records that are delimited by a specific start/end tag.
```
srowen: I'd revert these whitespace changes; we've used javadoc-style indentation like Spark does. No idea why Scala did it differently.

src/main/scala/com/databricks/spark/xml/XmlInputFormat.scala
```diff
@@ -25,14 +25,17 @@ import org.apache.hadoop.io.{DataOutputBuffer, LongWritable, Text}
...
+  */
 class XmlInputFormat extends TextInputFormat {
 
   override def createRecordReader(
-      split: InputSplit,
-      context: TaskAttemptContext): RecordReader[LongWritable, Text] = {
+                                   split: InputSplit,
```
srowen: Likewise this needs to be reverted

src/main/scala/com/databricks/spark/xml/XmlInputFormat.scala
```diff
@@ -189,9 +192,23 @@ private[xml] class XmlRecordReader extends RecordReader[LongWritable, Text] {
     false
   }
 
-  private def checkEmptyTag(currentLetter: Int, position: Int): Boolean = {
+  private def checkEmptyTag(currentLetter: Int, position: Int,
+                            buffer: DataOutputBuffer): Boolean = {
+    def checkStartTagBefore = {
+      val buf = Seq('<'.toByte)
+      val rootTagName = buffer.getData
```
srowen: Why define a method here?
I don't think this is efficient enough as it makes a few copies of much of the buffer.

src/main/scala/com/databricks/spark/xml/XmlInputFormat.scala
```diff
@@ -222,9 +239,9 @@ private[xml] class XmlRecordReader extends RecordReader[LongWritable, Text] {
             si += 1
             ei = 0
           }
-        } else if (b == endTag(ei) || checkEmptyTag(b, ei)) {
+        } else if ((b == endTag(ei)) || checkEmptyTag(b, ei, buffer)) {
```
srowen: The parens aren't needed

src/main/scala/com/databricks/spark/xml/XmlInputFormat.scala
BioQwer: I can't revert this change, because goal `scalacheckstyle` will fail on it.

image

src/main/scala/com/databricks/spark/xml/XmlInputFormat.scala
BioQwer: Because it's used only here.

> I don't think this is efficient enough as it makes a few copies of much of the buffer.

maybe, but this code run only in rare situations
code in buffet not so easy to read

srowen:
This should be fixed by https://github.com/databricks/spark-xml/pull/352

Issue #314, opened 2018-07-03
Cannot distinguish between empty and non-existent nodes


elibus:
Hi,
I have an XML as below:

```xml
<items>
	<Ntr>
		<Othr>NORE</Othr>
	</Ntr>
	<Ntr>
		<NFI>
		</NFI>
		<Othr>NORE</Othr>
	</Ntr>
	<Ntr>
		<NFI>
                    <VAL>some<VAL>
		</NFI>
		<Othr>NORE</Othr>
	</Ntr>
</items>
```

I can parse correctly the structure above, however I am not able to tell whether the NFI node is present and empty or is not present at all because, in both cases for the first and the second NFI elements above the parser returns NULL.

What I would expect as behavior is the parser to:

- return NULL if the NFI element does not exist at all. (Element 1 of the XML)
- return a complex empty struct if the element exists but is empty. (Element 2 of the XML)
- return a complex struct filled with one element

Would that be reasonable?

elibus:
Linked to #117

HyukjinKwon:
@elibus, does this happen in the master branch too?

elibus:
@HyukjinKwon we are using branch 0.3 however, if I am not mistaken, the code responsible for this behaviour should be this one: https://github.com/databricks/spark-xml/blob/f3772b296d77a5a92ae4e714f3f8cb9940f1f0f9/src/main/scala/com/databricks/spark/xml/parsers/StaxXmlParser.scala#L218 which is identical in both 0.3 and 0.4 branches and introduced to fix #117 .

elibus:
Hi,
below an example XML and an explanation of the expected behaviour.

```xml
<root>
  <item>
    <!-- Issue 314 - We expect [null]
               The node exists but is empty
    -->
    <b>
    </b>
  </item>
  <item>
    <b>
      <!-- Issue 314 - We expect [[null,null]]
                   The <es> node exists and can have two nested elements, both null
      -->
      <es></es>
    </b>
  </item>
  <item>
    <b>
      <es>
        <!-- Issue 314 - We expect [[E,null]]
                       The <es> node exists and can have two nested elements, E and null
        -->
        <e>E</e>
      </es>
    </b>
  </item>
  <item>
    <b>
      <es>
        <!-- Issue 314 - We expect [[E,null]]
                       The <es> node exists and can have two nested elements, E and " "
        -->
        <e>E</e>
        <f> </f>
      </es>
    </b>
  </item>
  <item>
    <b>
[...]

PR #325, opened 2018-08-13
[FIX] Returns an empty Row() when a node element is empty

elibus:
Fix for #314 .

In the XML below the node element `<es></es>` exists and is empty in the first item while does not exist in the second. In both cases the parser returns `null`.
This behaviour does not allow to distinguish when an optional node element exists but it is empty and when it is does not exist.

The patch preserves the behaviour of option `treatEmptyValuesAsNulls` by returning `null` if set to true.

```xml
<?xml version="1.0" ?>
<root>
	<item>
		<b>
			<!-- this one returns null -->
			<es></es>
                         <c>a value</c>
		</b>
	</item>
	<item>
		<b>
			<!-- this one returns also null for  <es></es> -->
                         <c>a value</c>
		</b>
	</item>
</root>
```

### DIFF
--- a/src/main/scala/com/databricks/spark/xml/parsers/StaxXmlParser.scala
+++ b/src/main/scala/com/databricks/spark/xml/parsers/StaxXmlParser.scala
@@ -218,7 +218,11 @@ private[xml] object StaxXmlParser {
     // Return null rather than empty row. For nested structs empty row causes
     // ArrayOutOfBounds exceptions when executing an action.
     if (valuesMap.isEmpty) {
-      null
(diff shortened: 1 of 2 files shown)


elibus:
I am closing this because I found out that it raises a java.lang.ArrayIndexOutOfBoundsException with a schema like the below. Removing the f element from the schema works as expected. 

```scala
    val customSchema = new StructType()
      .add("b",
        new StructType()
          .add("es", new StructType()
            .add("e", StringType)
            .add("f", StringType)
          )
        )
```

PR #326, opened 2018-08-14
Returns an empty Row instead of null when a node has no children

elibus:
See #314 for details

- Returns an empty Row instead of null when a node element has no children
- Returns an empty string instead of null when the node element has no next


srowen:
Thanks @elibus ! I merge a variant of this after rebasing your change https://github.com/databricks/spark-xml/pull/355

PR #330, opened 2018-09-08
Support Xml Boolean Lexical Literal 1 and 0 for boolean data type

skadyan:
- Added Support for supporting xsd boolean lexical literals 1, 0 along with true and false. 
- Specified encoding was getting ignored. 
- Added clause in the throws exception (in case of FAILFAST mode) to add root cause.

### DIFF
--- a/src/test/scala/com/databricks/spark/xml/XmlSuite.scala
+++ b/src/test/scala/com/databricks/spark/xml/XmlSuite.scala
@@ -112,6 +112,31 @@ final class XmlSuite extends FunSuite with BeforeAndAfterAll {
     Files.createTempDirectory(tempDir, "test")
   }
 
+  // Utilities to simplify schema specification:
+
+  private def buildSchema(fields: StructField*): StructType =
+    StructType(fields)
+
+  private def field(
+      name: String,
+      dataType: DataType = StringType,
+      nullable: Boolean = true): StructField =
+    StructField(name, dataType, nullable)
+
+  private def struct(fields: StructField*): StructType =
+    buildSchema(fields: _*)
+
+  private def struct(name: String, fields: StructField*): StructField =
+    field(name, struct(fields: _*))
+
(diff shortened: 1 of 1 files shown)


src/main/scala/com/databricks/spark/xml/util/TypeCast.scala
```diff
@@ -70,6 +70,18 @@ object TypeCast {
     }
   }
 
+  private def parseXmlBoolean(s: String): Boolean = {
```
srowen: This part looks reasonable but let's restrict the change to just this part.

src/main/scala/com/databricks/spark/xml/util/TypeCast.scala
```diff
@@ -70,6 +70,18 @@ object TypeCast {
...
+    if (s != null) s.toLowerCase match {
+      case "true" => true
+      case "false" => false
+      case "1" => true
+      case "0" => false
+      case _ => throw new IllegalArgumentException("For input string: \"" + s + "\"")
+    }
+    else {
+      throw new IllegalArgumentException("For input string: \"null\"")
```
srowen: Skip this clause and just `s.toLowerCase match`. It will throw an NPE as before.

src/main/scala/com/databricks/spark/xml/util/TypeCast.scala
```diff
@@ -70,6 +70,18 @@ object TypeCast {
     }
   }
 
+  private def parseXmlBoolean(s: String): Boolean = {
+    if (s != null) s.toLowerCase match {
+      case "true" => true
+      case "false" => false
+      case "1" => true
+      case "0" => false
+      case _ => throw new IllegalArgumentException("For input string: \"" + s + "\"")
```
srowen: Use string interpolation here

HyukjinKwon:
@skadyan Looks something went wrong during updating this PR. Mind rebaseing it or making a PR again?

Issue #322, opened 2018-08-06
UDF to parse a column with an XML string


stevenmanton:
I love this package, but I have often run into a scenario where I have a DataFrame with several columns, one of which contains an XML string that I would like to parse. Since this package only works with files, in order to parse the XML column we have to select the XML column, save it to disk, then read it using this library.

I'd love a UDF that I could call that would parse the column in place. For example, a new function `parseXML` that parses the XML string and returns a struct that you could reference in the normal way. Maybe something along the lines of the following.
```python
(
    df
    .withColumn("parsed_XML", parseXML('xml_column'))
    .withColumn("field1", "parsed_XML.field1")
    .withColumn("array0", col("parsed_XML.array").getItem(0))
)
```

I'm happy to try to implement this, but I'm hoping the core devs can provide some early feedback. Is this doable? worthwhile? Any suggestions on the right approach?

HyukjinKwon:
If you are willing to make a implementation, sounds like a good idea (like from_json and to_json function implementations in Spark side).

Should be good to refer https://github.com/apache/spark/pull/21838.

SpaceRangerWes:
@stevenmanton @HyukjinKwon I attempted an initial implementation of this enhancement. Love to work with you to fulfill any requirements.

srowen:
Yes this is a duplicate of a few other issues, like https://github.com/databricks/spark-xml/pull/334

PR #334, opened 2018-10-29
Implementation of from_xml function

SpaceRangerWes:
Issue #322 wanted an implementation of a commonly asked for feature. The purpose of this PR is to provide a new function `from_xml` that leverages as much of the current code as possible to parse a string encoded XML column as a complex DataType when provided a schema.

Both fortunately and unfortunately, my team at work is pivoting away from any nested XML objects so I was not allowed more time to provide a larger set of enhancements. I'd gladly make any small changes requested, but I will not have the capacity to provide any further enhancements.


SpaceRangerWes:
Note on the build fail:

This is occurring for Scala: 2.10.5 JDK: openjdk7 on a preexisting dependency. I don't know of any of my code that would utilize or break this dependency. Therefore, any dependency tree management is out of scope for this task.

hina1234:
Hi,

I have the similar requirement, where I need to do analysis on a column with xml string. Will be great if you could provide an example on how to use this .

SpaceRangerWes:
@hina1234 I added a unit test that shows an example of using what I have written. Looking at it now, the test is poorly name, but it should be what you are looking for https://github.com/SpaceRangerWes/spark-xml/blob/master/src/test/scala/com/databricks/spark/xml/XmlSuite.scala#L919-L941

HyukjinKwon:
Sounds good to have it. Can you rebase and make the tests pass?

SpaceRangerWes:
@HyukjinKwon finally rebased. Sorry that it took so long. The holidays were the only time I had available.

src/main/scala/com/databricks/spark/xml/XmlOptions.scala
```diff
@@ -51,6 +51,7 @@ private[xml] class XmlOptions(
   val ignoreSurroundingSpaces =
     parameters.get("ignoreSurroundingSpaces").map(_.toBoolean).getOrElse(false)
   val parseMode = ParseMode.fromString(parameters.getOrElse("mode", PermissiveMode.name))
+  val isFunction: Boolean = parameters.get("isFunction").exists(_.toBoolean)
```
HyukjinKwon: Looks it's not used. Can we remove this?

src/main/scala/com/databricks/spark/xml/package.scala
```diff
@@ -16,13 +16,47 @@
 package com.databricks.spark
 
 import scala.collection.Map
-
 import org.apache.hadoop.io.compress.CompressionCodec
-
```
HyukjinKwon: Actually this project follows the style guide here https://github.com/databricks/scala-style-guide#imports. Should be good to match the import order.

src/main/scala/com/databricks/spark/xml/XmlDataToCatalyst.scala
```diff
@@ -0,0 +1,44 @@
...
+      s"Input schema ${schema.simpleString} must be a struct or an array of structs.")
+  }
+
+  @transient
+  lazy val rowSchema: StructType = schema match {
+    case st: StructType => st
+    case ArrayType(st: StructType, _) => st
+  }
+
+  override def nullSafeEval(xml: Any): Any = {
```
HyukjinKwon: nit:

```scala
override def nullSafeEval(xml: Any): Any = xml match {
  case ...
}
```

src/main/scala/com/databricks/spark/xml/XmlDataToCatalyst.scala
```diff
@@ -0,0 +1,44 @@
...
+import org.apache.spark.sql.catalyst.CatalystTypeConverters
+import org.apache.spark.sql.catalyst.analysis.TypeCheckResult
+import org.apache.spark.sql.catalyst.expressions.codegen.CodegenFallback
+import org.apache.spark.sql.catalyst.expressions.{ExpectsInputTypes, Expression, UnaryExpression}
+import org.apache.spark.sql.types._
+import org.apache.spark.unsafe.types.UTF8String
+
+case class XmlDataToCatalyst(child: Expression,
+                             schema: DataType,
+                             options: XmlOptions)
```
HyukjinKwon: nit:

```scala
case class XmlDataToCatalyst(
    child: Expression,
    schema: DataType,
    options: XmlOptions)
  extends ...
```

(per the guide I pointed out above)

src/main/scala/com/databricks/spark/xml/XmlDataToCatalyst.scala
```diff
@@ -0,0 +1,44 @@
...
+    case _: StructType | ArrayType(_: StructType, _) =>
+      super.checkInputDataTypes()
+    case _ => TypeCheckResult.TypeCheckFailure(
+      s"Input schema ${schema.simpleString} must be a struct or an array of structs.")
+  }
+
+  @transient
+  lazy val rowSchema: StructType = schema match {
+    case st: StructType => st
+    case ArrayType(st: StructType, _) => st
```
HyukjinKwon: I think XML implementation does not support an array of struct. Can be removed.

src/main/scala/com/databricks/spark/xml/parsers/StaxXmlParser.scala
```diff
@@ -111,6 +113,30 @@ private[xml] object StaxXmlParser extends Serializable {
     }
   }
 
+  def parseColumn(xml: String,
+                  schema: StructType,
+                  options: XmlOptions): Row = {
+    val factory: XMLInputFactory = XMLInputFactory.newInstance()
```
HyukjinKwon: Let's make a private function for:

```scala
    val factory: XMLInputFactory = XMLInputFactory.newInstance()
    factory.setProperty(XMLInputFactory.IS_NAMESPACE_AWARE, false)
    factory.setProperty(XMLInputFactory.IS_COALESCING, true)

    val filter = new EventFilter {
      override def accept(event: XMLEvent): Boolean =
      // Ignore comments. This library does not treat comments.
        event.getEventType != XMLStreamConstants.COMMENT
    }
```

and deduplicate the logic with `parse` above.

Also, I would make a function to deduplicate

```scala
    val eventReader = factory.createXMLEventReader(new StringReader(xml))
    val parser = factory.createFilteredReader(eventReader, filter)

    val rootEvent =
      StaxXmlParserUtils.skipUntil(parser, XMLStreamConstants.START_ELEMENT)
    val rootAttributes =
      rootEvent.asStartElement.getAttributes.asScala.map(_.asInstanceOf[Attribute]).toArray
    convertObject(parser, schema, options, rootAttributes)
```

as well.

src/main/scala/com/databricks/spark/xml/parsers/StaxXmlParser.scala
```diff
@@ -111,6 +113,30 @@ private[xml] object StaxXmlParser extends Serializable {
     }
   }
 
+  def parseColumn(xml: String,
+                  schema: StructType,
+                  options: XmlOptions): Row = {
```
HyukjinKwon: nit:

```
def parseColumn(
  xml: String,
  ...): Row = {
```

src/main/scala/com/databricks/spark/xml/parsers/StaxXmlParserUtils.scala
```diff
@@ -131,4 +133,20 @@ private[xml] object StaxXmlParserUtils {
       }
     }
   }
+
+  def createXmlEventReader(xml: Array[Byte]): XMLEventReader = {
```
HyukjinKwon: Looks not used.

src/test/scala/com/databricks/spark/xml/XmlSuite.scala
```diff
@@ -1009,4 +1012,97 @@ final class XmlSuite extends FunSuite with BeforeAndAfterAll {
       <integer_value int="Ten">Ten</integer_value>.toString))
   }
 
+
+  test("from_xml roundtrip happy path") {
+
+    val xmlData =
+      s"""
```
HyukjinKwon: `s` looks not needed.

src/test/scala/com/databricks/spark/xml/XmlSuite.scala
```diff
@@ -1009,4 +1012,97 @@ final class XmlSuite extends FunSuite with BeforeAndAfterAll {
...
+    val expectedSchema: StructType = rowSchema.add("decoded", xmlSchema)
+
+    val df: DataFrame = spark.createDataFrame(
+      spark.sparkContext.parallelize(List(Row(8, xmlData))), rowSchema)
+
+    val result: DataFrame = df.withColumn("decoded",
+      from_xml(df.col("payload"), xmlSchema))
+
+    assert(expectedSchema == result.schema)
+    assert(result.where(col("decoded").isNotNull).count() > 0)
```
HyukjinKwon: I think the actual output was also asserted and checked for completeness.

src/test/scala/com/databricks/spark/xml/XmlSuite.scala
```diff
@@ -1009,4 +1012,97 @@ final class XmlSuite extends FunSuite with BeforeAndAfterAll {
...
+  }
+
+  /*  this unit test shows that there is no current support for changing the root tag
+     when using this function. Support may be added later. In which case, this test should pass */
+  test("from_xml roundtrip with rowTag or rootTag fails") {
+
+    val xmlData =
+      s"""
```
HyukjinKwon: `s` doesn't looks needed as well.

src/test/scala/com/databricks/spark/xml/XmlSuite.scala
```diff
@@ -1009,4 +1012,97 @@ final class XmlSuite extends FunSuite with BeforeAndAfterAll {
...
+    val result: DataFrame = df.withColumn("decoded",
+      from_xml(df.col("payload"), xmlSchema))
+
+    assert(expectedSchema == result.schema)
+    assert(result.where(col("decoded").isNotNull).count() > 0)
+  }
+
+  /*  this unit test shows that there is no current support for changing the root tag
+     when using this function. Support may be added later. In which case, this test should pass */
+  test("from_xml roundtrip with rowTag or rootTag fails") {
```
HyukjinKwon: It's okie to remove this test ... don't worry too much :-).

src/test/scala/com/databricks/spark/xml/XmlSuite.scala
```diff
@@ -1009,4 +1012,97 @@ final class XmlSuite extends FunSuite with BeforeAndAfterAll {
...
+
+    assert(expectedSchema == result.schema)
+    assert(result.where(col("decoded").isNotNull).count() > 0)
+
+    assert(0 == result
+      .where(col("decoded.pid").isNotNull or col("decoded.name").isNotNull).count())
+  }
+
+
+  test("from_xml array support") {
```
HyukjinKwon: Let's don't add array support for now :D. JSON in Spark added the support for an array a long time ago .. and it's there IMHO for legacy reason. To be honest, I doubt it was a great idea yet.

src/main/scala/com/databricks/spark/xml/XmlDataToCatalyst.scala
```diff
@@ -0,0 +1,44 @@
+package com.databricks.spark.xml
```
HyukjinKwon: Looks license missing. We can copy and paste the lisence from another file.

HyukjinKwon:
Thanks, @SpaceRangerWes for working on this!

I took a quick pass. Should be good to refer https://github.com/databricks/scala-style-guide for code style.

src/main/scala/com/databricks/spark/xml/XmlDataToCatalyst.scala
SpaceRangerWes: @HyukjinKwon I'm having a challenging time getting intellij to enforce this. I'm going to try swinging back around to it after I commit the more sizable change requests.

src/main/scala/com/databricks/spark/xml/XmlDataToCatalyst.scala
SpaceRangerWes: Would you like me to have a `case _ => throw SomeExceptionClass('string')` statement? Otherwise `rowSchema` must be of type `DataType`

src/main/scala/com/databricks/spark/xml/package.scala
SpaceRangerWes: Same with https://github.com/databricks/spark-xml/pull/334/files/ca711d7da1762bc88c48f1dea9fadc5b0b727c83#r244904273

src/main/scala/com/databricks/spark/xml/XmlDataToCatalyst.scala
HyukjinKwon: Ah, I think we can change `DataType` to `StructType`.

src/main/scala/com/databricks/spark/xml/XmlDataToCatalyst.scala
HyukjinKwon: Yea, it is a bit grunting to manually fix the style. I admit. Such change suggestion wouldn't block this PR but should be good to follow the style guide (https://github.com/databricks/scala-style-guide) since Databricks repositories comply this style .. (BTW, I'm not a Databricks guy so I would like to suggest follow this guide to be safe).

src/main/scala/com/databricks/spark/xml/parsers/StaxXmlParser.scala
```diff
@@ -82,25 +106,11 @@ private[xml] object StaxXmlParser extends Serializable {
...
-          // Ignore comments. This library does not treat comments.
-          event.getEventType != XMLStreamConstants.COMMENT
-      }
 
       iter.flatMap { xml =>
-        // It does not have to skip for white space, since `XmlInputFormat`
-        // always finds the root tag without a heading space.
-        val eventReader = factory.createXMLEventReader(new StringReader(xml))
-        val parser = factory.createFilteredReader(eventReader, filter)
+        val parser = filteredReader(xml)
```
HyukjinKwon: This will called for every record. We should make the factory for each partition like the previous code, and make each reader for each record. Such pattern of computation per record causes a considerable performance regression.

SpaceRangerWes: @HyukjinKwon I think I've found a semi-simple solution in the latest commit. Look good?

src/main/scala/com/databricks/spark/xml/parsers/StaxXmlParser.scala
```diff
@@ -111,6 +123,20 @@ private[xml] object StaxXmlParser extends Serializable {
     }
   }
 
+  def parseColumn(xml: String,
+                  schema: StructType,
+                  options: XmlOptions): Row = {
+    val factory: XMLInputFactory = XMLInputFactory.newInstance()
```
HyukjinKwon: Currently, 

1. XML string
2. XMLInputFactory instance is created
3. XML is parsed to `Row`

I think this can be avoided by:

```scala
class XmlDataToCatalyst(..) {
   @transient
  lazy val factory: XMLInputFactory = XMLInputFactory.newInstance()

  def nullSafeEval(...) = {
    factory.createXMLEventReader(new StringReader(xml))
  }
}
```

In this way, we will create one factory first, and then reuse it for every string input. (BTW, it's quite critical to fix this.).

HyukjinKwon:
BTW, @SpaceRangerWes, why don't you address https://github.com/databricks/spark-xml/pull/334#discussion_r244669868, https://github.com/databricks/spark-xml/pull/334#discussion_r244669715 and https://github.com/databricks/spark-xml/pull/334#discussion_r244669651 while addressing other comments?

Strictly, some codes look not complying the style that this project follow (https://github.com/databricks/scala-style-guide).

mantovani:
Can I do something to help ? I really need this feature.

srowen:
@mantovani you can open a new PR with these changes and address the comments maybe?

SpaceRangerWes:
@srowen @mantovani I really apologize. I just don't have the capacity to continue the work on this and have had to abandon it. My team's needs migrated and I could no longer work on this.

PR #343, opened 2018-12-17
Fix minor issues to enable Scala 2.12, Spark 2.4 supports. Update plugins too

srowen:
CC @HyukjinKwon -- this might be a good prelude to further fixes and a new release. This should make sure Java 8, Spark 2.4 and Scala 2.12 work. Along the way I suggest we forget about Spark before 2.2. I updated plugins to match Spark for good measure.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,6 +1,6 @@
 name := "spark-xml"
 
-version := "0.4.2"
+version := "0.5.0"
 
 organization := "com.databricks"
 
@@ -15,12 +15,9 @@ sparkVersion := sys.props.get("spark.testVersion").getOrElse("2.4.0")
 sparkComponents := Seq("core", "sql")
 
 libraryDependencies ++= Seq(
-  "org.slf4j" % "slf4j-api" % "1.7.5" % "provided",
+  "org.slf4j" % "slf4j-api" % "1.7.25" % "provided",
   "org.scalatest" %% "scalatest" % "3.0.3" % "test",
-  "com.novocode" % "junit-interface" % "0.9" % "test"
-)
-
-libraryDependencies ++= Seq(
+  "com.novocode" % "junit-interface" % "0.11" % "test",
   "org.apache.spark" %% "spark-core" % sparkVersion.value % "test",
   "org.apache.spark" %% "spark-sql" % sparkVersion.value % "test",
   "org.scala-lang" % "scala-library" % scalaVersion.value % "compile"
@@ -36,7 +33,7 @@ spAppendScalaVersion := true
 
 spIncludeMaven := true
 
-pomExtra := (
+pomExtra :=
   <url>https://github.com/databricks/spark-xml</url>
   <licenses>
     <license>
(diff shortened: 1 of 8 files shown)


build.sbt
```diff
@@ -4,34 +4,25 @@ version := "0.4.2"
...
 
-scalaVersion := "2.11.7"
+scalaVersion := "2.11.12"
 
 spName := "databricks/spark-xml"
 
-crossScalaVersions := Seq("2.10.5", "2.11.7")
+crossScalaVersions := Seq("2.11.12", "2.12.8")
 
-sparkVersion := "2.0.0"
```
srowen: I'm not sure about this, but this seemed to hard-code Spark versions to 2.0.0 despite the command line setting.

build.sbt
```diff
@@ -4,34 +4,25 @@ version := "0.4.2"
...
-
-val testSparkVersion = settingKey[String]("The version of Spark to test against.")
-
-testSparkVersion := sys.props.get("spark.testVersion").getOrElse(sparkVersion.value)
-
-val testHadoopVersion = settingKey[String]("The version of Hadoop to test against.")
```
srowen: I believe we can just assume Hadoop 2.7 for Spark 2.2+ here.

build.sbt
```diff
@@ -4,34 +4,25 @@ version := "0.4.2"
...
 
 libraryDependencies ++= Seq(
   "org.slf4j" % "slf4j-api" % "1.7.5" % "provided",
-  "org.scalatest" %% "scalatest" % "2.2.1" % "test",
+  "org.scalatest" %% "scalatest" % "3.0.3" % "test",
   "com.novocode" % "junit-interface" % "0.9" % "test"
 )
 
 libraryDependencies ++= Seq(
-  "org.apache.hadoop" % "hadoop-client" % testHadoopVersion.value % "test",
```
srowen: ... and just use whatever Spark brings

src/main/scala/com/databricks/spark/xml/parsers/StaxXmlParser.scala
```diff
@@ -35,7 +35,7 @@ import com.databricks.spark.xml.XmlOptions
 /**
  * Wraps parser to iteration process.
  */
-private[xml] object StaxXmlParser {
+private[xml] object StaxXmlParser extends Serializable {
```
srowen: Not clear why, but this caused a bunch of lambdas to not serialize in 2.12. It should be harmless to let the object serialize; it has no fields.

src/test/scala/com/databricks/spark/xml/XmlSuite.scala
```diff
@@ -207,26 +207,23 @@ class XmlSuite extends FunSuite with BeforeAndAfterAll {
   test("DDL test") {
     sqlContext.sql(
       s"""
-         |CREATE TEMPORARY TABLE carsTable
+         |CREATE TEMPORARY TABLE carsTable1
```
srowen: This started failing in Spark 2.3 as the temp table already existed across examples.

srowen:
I'm going to merge this as it fixes some problems and unblocks a number of other changes we need to test

PR #346, opened 2018-12-18
Use SparkSession in favor of deprecated SQLContext in tests (deprecation warning)

srowen:
Use SparkSession in favor of deprecated SQLContext in tests (deprecation warning), and add method that directly accepts SparkSession. Plus minor fixes elsewhere to the code that raise IJ warnings.

### DIFF
--- a/src/main/scala/com/databricks/spark/xml/XmlReader.scala
+++ b/src/main/scala/com/databricks/spark/xml/XmlReader.scala
@@ -92,14 +92,38 @@ class XmlReader extends Serializable {
     this
   }
 
+  def xmlFile(spark: SparkSession, path: String): DataFrame = {
+    // We need the `charset` and `rowTag` before creating the relation.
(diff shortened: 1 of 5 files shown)


src/main/scala/com/databricks/spark/xml/XmlReader.scala
```diff
@@ -116,4 +115,11 @@ class XmlReader extends Serializable {
       schema)(sqlContext)
     sqlContext.baseRelationToDataFrame(relation)
   }
+
+  def xmlFile(spark: SparkSession, path: String): DataFrame =
```
HyukjinKwon: I was actually thinking about deprecating these too as well. but I think we can do later.

srowen:
I'll deprecate those other methods, I agree.

HyukjinKwon:
Thank you @srowen for tacking a look for all those stuff.

Issue #120, opened 2016-05-10
[Bug] XmlInputFormat doesn't play nice with newlines


maarest:
Hi,

I have no idea how to fix it but this seems a pretty tricky bug

The following code:
```
    // This will detect the tags including attributes
    sc.hadoopConfiguration.set(XmlInputFormat.START_TAG_KEY, s"<note>")
    sc.hadoopConfiguration.set(XmlInputFormat.END_TAG_KEY, s"</note>")
    sc.hadoopConfiguration.set(XmlInputFormat.ENCODING_KEY, "utf-8")

    sc.newAPIHadoopFile(path,classOf[XmlInputFormat],classOf[LongWritable],classOf[Text])
```

works for the following xml:
```
<?xml version="1.0"?>

<note xmlns="http://www.w3schools.com"
xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
xsi:schemaLocation="http://www.w3schools.com note.xsd">
  <to>Tove</to>
  <from>Jani</from>
  <heading>Reminder</heading>
  <body>Don't forget me this weekend!</body>
</note>
```
but doesn't work for the following xml (newline right after tag)
```
<?xml version="1.0"?>

<note
xmlns="http://www.w3schools.com"
xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
xsi:schemaLocation="http://www.w3schools.com note.xsd">
  <to>Tove</to>
  <from>Jani</from>
  <heading>Reminder</heading>
  <body>Don't forget me this weekend!</body>
</note>
```

HyukjinKwon:
Hm.. I see. I will try to reproduce this. Thank you for reporting this!

HyukjinKwon:
Indeed. This should be around here https://github.com/databricks/spark-xml/blob/master/src/main/scala/com/databricks/spark/xml/XmlInputFormat.scala#L213-L225.

PR #350, opened 2018-12-19
Check for whitespace after tag including newlines to signal tag with attribute

srowen:
See also https://github.com/databricks/spark-xml/pull/342
I also believe this resolves https://github.com/databricks/spark-xml/issues/120


BioQwer:
> rewrite the XmlInputFormat to make it not crazy hard to fix some other issues

waiting for it 👍

PR #355, opened 2018-12-21
Return Row of null instead of null for empty elements that are structs in schema

srowen:
This is a rebase and modification of the fix at https://github.com/databricks/spark-xml/pull/326 which addressed https://github.com/databricks/spark-xml/issues/314 both from @elibus .

It's a behavior change but I think the behavior change is correct, and was only inadvertently introduced in https://github.com/databricks/spark-xml/pull/118

### DIFF
--- a/src/test/scala/com/databricks/spark/xml/XmlSuite.scala
+++ b/src/test/scala/com/databricks/spark/xml/XmlSuite.scala
@@ -35,38 +35,40 @@ import org.apache.spark.SparkException
 
 final class XmlSuite extends FunSuite with BeforeAndAfterAll {
 
-  private val agesFile = "src/test/resources/ages.xml"
-  private val agesWithSpacesFile = "src/test/resources/ages-with-spaces.xml"
-  private val booksFile = "src/test/resources/books.xml"
-  private val booksNestedObjectFile = "src/test/resources/books-nested-object.xml"
-  private val booksNestedArrayFile = "src/test/resources/books-nested-array.xml"
-  private val booksComplicatedFile = "src/test/resources/books-complicated.xml"
+  private val resDir = "src/test/resources/"
+  private val agesFile = resDir + "ages.xml"
+  private val agesWithSpacesFile = resDir + "ages-with-spaces.xml"
+  private val booksFile = resDir + "books.xml"
+  private val booksNestedObjectFile = resDir + "books-nested-object.xml"
+  private val booksNestedArrayFile = resDir + "books-nested-array.xml"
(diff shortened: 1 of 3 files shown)


src/test/scala/com/databricks/spark/xml/XmlSuite.scala
```diff
@@ -35,38 +35,40 @@ import org.apache.spark.SparkException
...
 final class XmlSuite extends FunSuite with BeforeAndAfterAll {
 
-  private val agesFile = "src/test/resources/ages.xml"
-  private val agesWithSpacesFile = "src/test/resources/ages-with-spaces.xml"
-  private val booksFile = "src/test/resources/books.xml"
-  private val booksNestedObjectFile = "src/test/resources/books-nested-object.xml"
-  private val booksNestedArrayFile = "src/test/resources/books-nested-array.xml"
-  private val booksComplicatedFile = "src/test/resources/books-complicated.xml"
+  private val resDir = "src/test/resources/"
+  private val agesFile = resDir + "ages.xml"
```
srowen: Most of the change in this file was further standardizing of the test code

src/test/scala/com/databricks/spark/xml/XmlSuite.scala
```diff
@@ -722,14 +717,35 @@ final class XmlSuite extends FunSuite with BeforeAndAfterAll {
...
-  test("Missing nested struct represented as null instead of empty Row") {
+  test("Missing nested struct represented as Row of nulls instead of null") {
     val result = spark.read
       .option("rowTag", "item")
       .xml(nullNestedStructFile)
       .select("b.es")
       .collect()
 
-    assert(result(1).toSeq === Seq(null))
+    assert(result(1).getStruct(0) !== null)
```
srowen: These changes represent the new behavior, tested by tests

src/test/scala/com/databricks/spark/xml/XmlSuite.scala
```diff
@@ -834,13 +850,17 @@ final class XmlSuite extends FunSuite with BeforeAndAfterAll {
     val resultsOne = spark.read
       .option("treatEmptyValuesAsNulls", "true")
       .xml(gpsEmptyField)
-    assert(resultsOne.selectExpr("extensions.TrackPointExtension").head().isNullAt(0))
+    assert(resultsOne.selectExpr("extensions.TrackPointExtension").head.getStruct(0) !== null)
```
srowen: These changes represent the new behavior as well

PR #358, opened 2018-12-22
Partially fill out row when invalid data is encountered

srowen:
This is a rebase and minor update of https://github.com/databricks/spark-xml/pull/294 and should resolve https://github.com/databricks/spark-xml/issues/235 , per @Woffendm


srowen:
@HyukjinKwon I noticed one thing about this just now -- in this case it doesn't add this record as a corrupted record. It also didn't before. Should it? I might be able to fix that quickly in a follow up... save any exceptions thrown while converting the object but continue, and throw the first one that occurred, to make it a failedRecord in `parse`?

HyukjinKwon:
Yea, it's okay. If that's not difficult, yes. we should fill the corrupt record.

srowen:
See https://github.com/databricks/spark-xml/pull/368 -- disregard my comment about corrupt rows; we wouldn't set those in permissive mode.

PR #364, opened 2018-12-22
Remove useless extending TableScan

HyukjinKwon:
We don't need to extend `TableScan` if it implements `PrunedScan`. See https://github.com/apache/spark/blob/5ad03607d1487e7ab3e3b6d00eef9c4028ed4975/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSourceStrategy.scala#L271-L302


srowen:
I don't know this part of Spark, so if you believe so that's fine. Just need to add the MiMa excludes of course.

PR #366, opened 2018-12-22
Add test utility methods to make schemas easier to write/read

srowen:


src/test/scala/com/databricks/spark/xml/XmlSuite.scala
```diff
@@ -112,6 +113,36 @@ final class XmlSuite extends FunSuite with BeforeAndAfterAll {
     Files.createTempDirectory(tempDir, "test")
   }
 
+  // Utilities to simplify schema specification:
+
+  @varargs
```
HyukjinKwon: @srowen, I think we don't need `varargs`. I think we can add it later when `JavaXmlSuite` actually uses it.

srowen: That's fine either way, I'll remove it. The Java tests don't need it right now.

PR #367, opened 2018-12-22
Support Xml Boolean Lexical Literal 1 and 0 for boolean data type

skadyan:
Added Support for supporting xsd boolean lexical literals 1, 0 along with true and false.
Added clause in the throws exception (in case of FAILFAST mode) to add root cause.

### DIFF
--- a/src/main/scala/com/databricks/spark/xml/parsers/StaxXmlParser.scala
+++ b/src/main/scala/com/databricks/spark/xml/parsers/StaxXmlParser.scala
@@ -44,14 +44,15 @@ private[xml] object StaxXmlParser extends Serializable {
       xml: RDD[String],
       schema: StructType,
       options: XmlOptions): RDD[Row] = {
-    def failedRecord(record: String): Option[Row] = {
+    def failedRecord(record: String, cause: Throwable = null): Option[Row] = {
       // create a row even if no corrupt record column is present
       options.parseMode match {
         case FailFastMode =>
           throw new RuntimeException(
-            s"Malformed line in FAILFAST mode: ${record.replaceAll("\n", "")}")
+            s"Malformed line in FAILFAST mode: ${record.replaceAll("\n", "")}", cause)
         case DropMalformedMode =>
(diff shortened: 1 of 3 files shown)


src/test/scala/com/databricks/spark/xml/util/TypeCastSuite.scala
```diff
@@ -68,6 +68,9 @@ final class TypeCastSuite extends FunSuite {
     assert(TypeCast.castTo("1.00", FloatType, options) === 1.0)
     assert(TypeCast.castTo("1.00", DoubleType, options) === 1.0)
     assert(TypeCast.castTo("true", BooleanType, options) === true)
+    assert(TypeCast.castTo("1", BooleanType, options) == true)
```
srowen: Tiny nit: use `===` to trigger the scalatest macros that print better errors when this fails

src/main/scala/com/databricks/spark/xml/parsers/StaxXmlParser.scala
```diff
@@ -44,14 +44,15 @@ private[xml] object StaxXmlParser extends Serializable {
...
+    def failedRecord(record: String, cause: Throwable = null): Option[Row] = {
       // create a row even if no corrupt record column is present
       options.parseMode match {
         case FailFastMode =>
           throw new RuntimeException(
-            s"Malformed line in FAILFAST mode: ${record.replaceAll("\n", "")}")
+            s"Malformed line in FAILFAST mode: ${record.replaceAll("\n", "")}", cause)
         case DropMalformedMode =>
-          logger.warn(s"Dropping malformed line: ${record.replaceAll("\n", "")}")
+          val reason = if (cause != null) cause.getMessage else "Unknown";
```
srowen: Tiny nit: You could print either "Reason: ${cause.getMessage}" or nothing at all instead of printing "Reason: Unknown" in the latter case

src/test/scala/com/databricks/spark/xml/util/TypeCastSuite.scala
skadyan: Fixed.

PR #368, opened 2018-12-22
Alternate take on continuing parsing rows in permissive mode

srowen:
See also https://github.com/databricks/spark-xml/pull/358

Alternate take on continuing parsing rows in permissive mode; add test for Map and don't add null array/map entry where invalid.

Much of the change is minor cleanup while trying to debug; the `shouldStop = shouldStop &&` conditions were always no-ops for instance.

I added a Map field to the test case.

In the end it's a more complex but maybe slightly more robust change as it handles errors in more places. It does change the answers slightly. In the case of an array field parsing:

```
        <integer_array>Ten</integer_array>
        <integer_array>2</integer_array>
```

The answer is now an array `[2]` not `[null, 2]`. I think that makes a little more sense. Fields that aren't parsed are `null` because something has to be its value. But adding an entry that's `null` to represent something that's invalid seems a little worse than adding nothing.

CC @Woffendm

### DIFF
--- a/src/main/scala/com/databricks/spark/xml/parsers/StaxXmlParser.scala
+++ b/src/main/scala/com/databricks/spark/xml/parsers/StaxXmlParser.scala
@@ -110,7 +110,7 @@ private[xml] object StaxXmlParser extends Serializable {
 
     (parser.peek, dataType) match {
       case (_: StartElement, dt: DataType) => convertComplicatedType(dt)
-      case (_: EndElement, dt: StringType) =>
+      case (_: EndElement, _: StringType) =>
         if (options.treatEmptyValuesAsNulls){
           null
         } else {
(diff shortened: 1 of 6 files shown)


src/main/scala/com/databricks/spark/xml/parsers/StaxXmlParser.scala
```diff
@@ -158,12 +158,15 @@ private[xml] object StaxXmlParser extends Serializable {
     while (!shouldStop) {
       parser.nextEvent match {
         case e: StartElement =>
-          keys += e.getName.getLocalPart
-          values += convertField(parser, valueType, options)
+          try {
+            keys += e.getName.getLocalPart
+            values += convertField(parser, valueType, options)
+          } catch {
+            case NonFatal(_) if options.parseMode == PermissiveMode => // do nothing
```
srowen: These try-catch blocks are the real change, in addition to undoing the change that handles all casts errors as null

srowen:
Actually I'm still a little unsure about what should happen here with the `columnNameOfCorruptRecord`. In a case where these sorts of parse errors are encountered, should we fill out the column? It didn't seem to do so before https://github.com/databricks/spark-xml/pull/358 and the change there changed docs to say it wasn't set. But should it be?

At least, I think this version of the change makes it easier to propagate an exception that occurs up to `convertObject` and `failedRow` if needed.

HyukjinKwon:
Hm, if it fails, the original record will be filled in the `columnnameOfCorruptRecord` via here:

https://github.com/databricks/spark-xml/blob/5e0fd2b9f6bd3950e5b37db62ac7722ff6e9e0b2/src/main/scala/com/databricks/spark/xml/parsers/StaxXmlParser.scala#L47-L96

PR #370, opened 2018-12-28
Produces partial results and also fills corrupt record

HyukjinKwon:
Followup for #368.

This PR proposes to produces partial results and also fills corrupt record. Some codes changes were borrowed from the resent change in Apache Spark (https://github.com/apache/spark/commit/4e1d859c19d3bfdfcb8acf915a97c68633b9ca95). That fix allow partial fields in JSON.

This PR fixes:
- Don't allow partial results within array and map (to match with Apache side)
- Partially parse and convert each value in each row. If it fails to parse or convert, it becomes `null`.
  - Partial results are only allowed in the top-level row. For instance,  nested partial result case like `Row(1, Row(1, 2, null))` is disallowed - it becomes `Row(1, null)`.
- If any exception is detected, the whole XML text is placed in `columnNameOfCorruptRecord` record

### DIFF
--- a/src/main/scala/com/databricks/spark/xml/parsers/StaxXmlParser.scala
+++ b/src/main/scala/com/databricks/spark/xml/parsers/StaxXmlParser.scala
@@ -24,6 +24,7 @@ import javax.xml.stream._
 import scala.collection.mutable.ArrayBuffer
 import scala.collection.JavaConverters._
 import scala.util.control.NonFatal
+import scala.util.Try
 
 import org.slf4j.LoggerFactory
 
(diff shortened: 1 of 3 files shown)


src/main/scala/com/databricks/spark/xml/parsers/StaxXmlParser.scala
```diff
@@ -44,7 +45,24 @@ private[xml] object StaxXmlParser extends Serializable {
       xml: RDD[String],
       schema: StructType,
       options: XmlOptions): RDD[Row] = {
-    def failedRecord(record: String, cause: Throwable = null): Option[Row] = {
+
+    // The logic below is borrowed from Apache Spark's FailureSafeParser.
```
HyukjinKwon: This logic below is borrowed from `FailureSafeParser` in Apache Spark.

src/main/scala/com/databricks/spark/xml/parsers/StaxXmlParser.scala
```diff
@@ -159,15 +173,12 @@ private[xml] object StaxXmlParser extends Serializable {
     while (!shouldStop) {
       parser.nextEvent match {
         case e: StartElement =>
```
HyukjinKwon: For Map and Arrow, it's not going to fill the partial results. It's matched to Apache Spark's JSON and actually quite complicated to support.

src/main/scala/com/databricks/spark/xml/parsers/StaxXmlParser.scala
```diff
@@ -260,51 +260,55 @@ private[xml] object StaxXmlParser extends Serializable {
...
         case _ => // do nothing
       }
     }
-    Row.fromSeq(row)
+
+    if (badRecordException.isEmpty) {
+      Row.fromSeq(row)
+    } else {
+      throw PartialResultException(Row.fromSeq(row), badRecordException.get)
+    }
```
HyukjinKwon: The main change in this PR is basically here:

```scala
var badRecordException: Option[Throwable] = None

while { // parse until the end
  try {
    row(i) = // parsed value
  } catch {
      case NonFatal(exception) if options.parseMode == PermissiveMode =>
        badRecordException = badRecordException.orElse(Some(exception))
  }
}

if (badRecordException.isEmpty) {
  Row.fromSeq(row)
} else {
  throw PartialResultException(Row.fromSeq(row), badRecordException.get)
}
```

src/main/scala/com/databricks/spark/xml/parsers/StaxXmlParser.scala
HyukjinKwon: Here's where `PartialResultException` is caught and handled.

PR #371, opened 2018-12-30
Make a release script and some fixes for Sonatype in build.sbt

HyukjinKwon:


HyukjinKwon:
BTW, I faced an issue that Scala version suffixes are missing. I think I always followed the current procedures so far if I am not mistaken so I suspect it's an issue somewhere.

For 0.5.0 release, I had to make some minor fixes in `build.sbt`.

I'm going to fix the script when 0.6.0 release is closed later if I need to fix something else.

srowen:
This makes sense. Yes, I was afraid that the several changes I made to the build script might have caused an issue at release time. I think the issue here is `crossPaths`. When `false`, it seems like the Java test suite is not run. Changing it to `true` looked like the recommended fix, but the flag really is about building separate outputs for different Scala versions. https://www.scala-sbt.org/1.0/docs/Basic-Def-Examples.html

If the issue is only that there is `target/scala` instead of `target/scala-2.11` then maybe the answer is to build separately for 2.11 and 2.12 or something.

Or you can just set this to `false` if it corrects the problem and we'll later find a different way to run the Java suite.

Or maybe that wasn't the problem you are seeing.

HyukjinKwon:
Yea, those stuff were what I faced. Let me get rid of `crossPaths` for now.

HyukjinKwon:
I just released it and tested this script.

build.sbt
```diff
@@ -10,9 +10,6 @@ spName := "databricks/spark-xml"
 
 crossScalaVersions := Seq("2.11.12", "2.12.8")
 
-// Necessary for JUnit tests to be found?
-crossPaths := false
```
HyukjinKwon: I don't think it was true .. this was ported from Spark CSV before.

I see Java tests are being ran fine:

```
[info] Test com.databricks.spark.xml.JavaXmlSuite.testLoad started
[info] Test com.databricks.spark.xml.JavaXmlSuite.testSave started
[info] Test com.databricks.spark.xml.JavaXmlSuite.testXmlParser started
```

HyukjinKwon: I at least see being ran fine:

```
[info] Test com.databricks.spark.xml.JavaXmlSuite.testLoad started
[info] Test com.databricks.spark.xml.JavaXmlSuite.testSave started
[info] Test com.databricks.spark.xml.JavaXmlSuite.testXmlParser started
```

srowen:
Ok that's good. Either I made a mistake or something else was actually the fix. Thanks!

Issue #246, opened 2017-03-20
More detailed error descriptions for unparsed files


someonehere15:
Is it possible to get more detailed descriptions for skipped or unparsed files?

It is as of now possible to get the corrupt column in the parsed DataFrame, or as a log entry if in DROPMALFORMED mode. But if a row is supposed to contain more complex structures (e.g. a directory with XML files and the rowTag is the root tag of each file, which would mean each XML file creates one row in the DF, and this row contains nested structures), then from the "_corrupt_column" column I can only get the problematic content not more granular than the rowTag (which would be a whole file in the example). So I still don't know which tag exactly was the problem inside of the defined rowTag, or what other errors there were.

Is there any way to extract more detailed informations on the actual cause of the problem than just the whole content of the problem row? At least a stack trace or similar should be possible to be retrieved at each moment that an error occurs, instead of simply skipping that row and putting its content in a column. Maybe by adding another column with an error description?

Yansongsongsong:
When I was parsing one xml encoded in `ISO-8859-1`,
the row with special character make some errors
e.g.:
```xml
<article mdate="2017-05-28" key="journals/acta/Schonhage77">
<author>Arnold Sch&ouml;nhage</author>
<title>Schnelle Multiplikation von Polynomen &uuml;ber K&ouml;rpern der Charakteristik 2.</title>
<journal>Acta Inf.</journal>
<volume>7</volume>
<year>1977</year>
<pages>395-398</pages>
<url>db/journals/acta/acta7.html#Schonhage77</url>
<ee>https://doi.org/10.1007/BF00289470</ee>
</article>
```
when i save the dataframe, i get the exception:
```scala
// code
selectedData.write
      .format("com.databricks.spark.xml")
      .option("charset", "ISO-8859-1")
      .option("rootTag", "articles")
      .option("rowTag", "article")
      .save(FILE_NAME + System.currentTimeMillis() + "article.xml")// get exception here
```

```
[ task launch worker for task 1] Utils                          ERROR Aborting task
java.lang.NullPointerException: null
	at com.databricks.spark.xml.parsers.StaxXmlGenerator$$anonfun$apply$4.apply(StaxXmlGenerator.scala:131) ~[spark-xml_2.11-0.4.1.jar:0.4.1]
[...]

PR #372, opened 2018-12-31
Log record parsing reasons at debug level in permissive, dropmalformed modes

srowen:


src/main/scala/com/databricks/spark/xml/parsers/StaxXmlParser.scala
```diff
@@ -66,13 +66,15 @@ private[xml] object StaxXmlParser extends Serializable {
       // create a row even if no corrupt record column is present
       options.parseMode match {
         case FailFastMode =>
-          throw new RuntimeException(
+          throw new IllegalArgumentException(
```
srowen: I also changed these to a more specific exception; anyone catching `RuntimeException` still catches this, but can maybe avoid catching every unchecked exception in the process.

src/main/scala/com/databricks/spark/xml/parsers/StaxXmlGenerator.scala
```diff
@@ -118,7 +118,7 @@ private[xml] object StaxXmlGenerator {
         }
 
       case (_, _) =>
-        sys.error(
+        throw new IllegalArgumentException(
```
srowen: OK, I replaced sys.error too as it generates just `RuntimeException`

srowen:
Closes https://github.com/databricks/spark-xml/issues/246

Issue #205, opened 2016-11-17
Disable inferSchema and keep everything as String


someonehere15:
Would it be possible to add an option to disable type detection and save all parsed values as a string datatype?

HyukjinKwon:
It is not. Thank you for a suggestion.

someonehere15:
So I just wanted to add a follow-up question here: Can I disable inferSchema, as is possible on spark-csv, and leave everything as a String?

jvmlet:
Have you tried to apply your schema explicitly ?

HyukjinKwon:
Yea, the best way is to give the schema explicitly. BTW. If we want to avoid only type inference, then it seems it goes in a problem in some cases like `inferSchema` option in CSV datasource.

someonehere15:
Same as OP of #223 , I also have a complex XML dataset and a dynamic schema on each run, so passing  a Schema explicitly is not an option for me. Creating a function as shown in #223 would require to load the files twice, which is a problem in my case because there are a lot of files. On the other hand, I would also need to adapt it for python, which I'm not sure exactly how to do it from the scala example (If anyone could help me there I could try it though despite the performance).

I understand it is going to be added, but in the meantime, is there anything I can change myself to the spark-xml source code to disable inferSchema or force everything to String, or is there going to be a new release soon?

HyukjinKwon:
I think we need this, at least, per use cases reported so far. I think it is not going to release soon though. Let me try to fix these and try to make a release at my best.

someonehere15:
Thank you very much, it would help me a lot.
I was thinking to change the inferSchema code to always return StringType or something like that, but if you could provide a better release, I would really appreciate it. Thanks again.

schoi99:
Just asking for a follow up, this thread has been open for a while. 
Is there any work around that could solve this issue temporally, or did I miss a release with a fix?
If any, some help would be greatly appreciated. Thank you

srowen:
You can just rewrite the schema to specify StringType in place of what is inferred, as a workaround. Read the file to infer the schema, then write a little code to map the primitive types to StringType, then reparse as string.

If the point is to avoid inference, yeah that won't help.
If the point is to just get string values out, you could call .toString on the values? cast them in the DataFrame?

blbradley:
@srowen That doesn't workfor columns that are integers but get inferred as doubles.

schoi99:
Yeah that was the issue. So here's my workaround. 

[...]

PR #393, opened 2019-06-13
Add support of custom column infer strategy #205

yjhyjhyjh0:
**Background**
Discussion of support all string type column inferring strategy.
Issues : https://github.com/databricks/spark-xml/issues/205

**Use case**
Want to use column inferring to take advantage of no predefined schema.
But want to treat every data as plain text instead of parsed one.

Ex: value “00010” will be inferred as long type and data becomes 10.
It’s not possible to recover data from 10 to “00010” during processing.

**Proposed Solution**
Add interface for custom column data type inferring strategy.

**Modification**
Add interface for column infer strategy.
Add new column infer strategy to treat every column as string.
Add unit test for default, all string infer strategy.

**Test result**
Pass scalastyle, all unit tests, local integration test.

Please let me know if there is any feedback on this pull request.
Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,7 @@ When reading files the API accepts several options:
     * When it encounters a field of the wrong datatype, it sets the offending field to `null`.
   * `DROPMALFORMED` : ignores the whole corrupted records.
   * `FAILFAST` : throws an exception when it meets corrupted records.
+* `inferSchema`: if `true`, attempts to infer an appropriate type for each resulting DataFrame column, like a boolean, numeric or date type. If `false`, all resulting columns are of string type. Default is `true`.
 * `columnNameOfCorruptRecord`: The name of new field where malformed strings are stored. Default is `_corrupt_record`.
 * `attributePrefix`: The prefix for attributes so that we can differentiate attributes and elements. This will be the prefix for field names. Default is `_`.
 * `valueTag`: The tag used for the value when there are attributes in the element having no child. Default is `_VALUE`.
--- a/src/main/scala/com/databricks/spark/xml/util/InferSchema.scala
+++ b/src/main/scala/com/databricks/spark/xml/util/InferSchema.scala
@@ -125,15 +125,19 @@ private[xml] object InferSchema {
       datum
     }
 
-    value match {
-      case null => NullType
-      case v if v.isEmpty => NullType
-      case v if isLong(v) => LongType
-      case v if isInteger(v) => IntegerType
-      case v if isDouble(v) => DoubleType
-      case v if isBoolean(v) => BooleanType
-      case v if isTimestamp(v) => TimestampType
-      case _ => StringType
+    if (options.inferSchema) {
+      value match {
+        case null => NullType
+        case v if v.isEmpty => NullType
+        case v if isLong(v) => LongType
+        case v if isInteger(v) => IntegerType
+        case v if isDouble(v) => DoubleType
+        case v if isBoolean(v) => BooleanType
+        case v if isTimestamp(v) => TimestampType
+        case _ => StringType
+      }
+    } else {
+      StringType
     }
   }
 
--- a/src/main/scala/com/databricks/spark/xml/XmlOptions.scala
+++ b/src/main/scala/com/databricks/spark/xml/XmlOptions.scala
@@ -51,6 +51,7 @@ private[xml] class XmlOptions(
   val ignoreSurroundingSpaces =
     parameters.get("ignoreSurroundingSpaces").map(_.toBoolean).getOrElse(false)
   val parseMode = ParseMode.fromString(parameters.getOrElse("mode", PermissiveMode.name))
+  val inferSchema = parameters.get("inferSchema").map(_.toBoolean).getOrElse(true)
 }
 
 private[xml] object XmlOptions {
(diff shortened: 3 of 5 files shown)


README.md
```diff
@@ -62,6 +62,9 @@ When reading files the API accepts several options:
     * When it encounters a field of the wrong datatype, it sets the offending field to `null`.
   * `DROPMALFORMED` : ignores the whole corrupted records.
   * `FAILFAST` : throws an exception when it meets corrupted records.
+* `inferStrategy`: The mode for dealing with corrupt records during parsing. Default is `PERMISSIVE`.
```
Wei-1: I think this should be:
```
    Default is `DEFAULT`.
```

yjhyjhyjh0: Wow, you're right.
I'll fix it now.

src/main/scala/com/databricks/spark/xml/util/InferSchema.scala
```diff
@@ -125,16 +125,7 @@ private[xml] object InferSchema {
...
-      case null => NullType
-      case v if v.isEmpty => NullType
-      case v if isLong(v) => LongType
-      case v if isInteger(v) => IntegerType
-      case v if isDouble(v) => DoubleType
-      case v if isBoolean(v) => BooleanType
-      case v if isTimestamp(v) => TimestampType
-      case _ => StringType
-    }
+    options.inferStrategy.getDataType(value)
```
HyukjinKwon: I think it doesn't need another object. Let's just do an if-else.

HyukjinKwon:
OK, let's add this option but I would name it `inferSchema`

src/main/scala/com/databricks/spark/xml/util/InferSchema.scala
yjhyjhyjh0: I was thinking about the same thing.
Since we only have two option for now, if-else would be enough.
Just add the interface to better explain the use case as the first draft solution.
I'm fine for both solution.
Please let me know if modification is required.

yjhyjhyjh0:
Hi @HyukjinKwon,
Thanks for the review.
I've modify the option name to “inferSchema”.

src/main/scala/com/databricks/spark/xml/util/InferSchema.scala
HyukjinKwon: Let's make it boolean and add an if-else. I have observed the usecases in Spark CSV datasource so far, and I think two cases are good enough.

srowen: Yeah, I can't think of another use case:
- User specifies schema -> use that
- User asks for inferred schema -> infer types
- Otherwise use the string rep that's already in the XML

That's how CSV works too, so being consistent and using a boolean here may be fine and less surprising.

So, what happens right now if you don't specify a schema? always inferred?

HyukjinKwon: Yea, it always infers ..

yjhyjhyjh0: Thanks for organizing the use cases.
I've pushed a version that uses if-else.

yjhyjhyjh0:
After some discussion of design and implementation. 
Just push new version that uses boolean and an if-else to indicate inferring or not.

Pass scalastyle, unit tests and local integration test.

HyukjinKwon:
Looks good. I'll merge this if there are no more comments.

src/main/scala/com/databricks/spark/xml/util/InferSchema.scala
```diff
@@ -125,15 +125,19 @@ private[xml] object InferSchema {
...
-    value match {
-      case null => NullType
-      case v if v.isEmpty => NullType
-      case v if isLong(v) => LongType
-      case v if isInteger(v) => IntegerType
-      case v if isDouble(v) => DoubleType
-      case v if isBoolean(v) => BooleanType
-      case v if isTimestamp(v) => TimestampType
-      case _ => StringType
+    if(options.inferSchema){
```
srowen: Not a big deal, but can we fix style here? spaces around parens.

README.md
```diff
@@ -62,6 +62,9 @@ When reading files the API accepts several options:
     * When it encounters a field of the wrong datatype, it sets the offending field to `null`.
   * `DROPMALFORMED` : ignores the whole corrupted records.
   * `FAILFAST` : throws an exception when it meets corrupted records.
+* `inferSchema`: Trying to infer column data type or not. Default is `true`.
```
srowen: I would just write...
```
`inferSchema`: if `true`, attempts to infer an appropriate type for each resulting DataFrame column, like a boolean, numeric or date type. If `false`, all resulting columns are of string type.
```

src/main/scala/com/databricks/spark/xml/util/InferSchema.scala
yjhyjhyjh0: Thanks for the reminder.
I'll fix it.

README.md
yjhyjhyjh0: That's indeed more concise.
I'll update the description.

yjhyjhyjh0:
Just pushed a new version for 
1- coding style. (spaces around parens)
2- update description on README.MD

Pass scalastyle, unit tests and local integration test.

HyukjinKwon:
Merged. Thanks.

PR #303, opened 2018-05-13
Add support for values with self-closing tags

HyukjinKwon:
Same for https://github.com/databricks/spark-xml/pull/285 but I resolved conflicts here. Should credit to him.

Closes #285 
Closes #156
Closes #92
Closes #190
Closes #276

PR #394, opened 2019-07-02
Self-closing tag support : no longer accurate

cefn:
There is a comment "At the moment, rows containing self closing xml tags are not supported." which may no longer be true as per... https://github.com/databricks/spark-xml/pull/303


HyukjinKwon:
Oh, nice!

ydalal927:
Does this change resolves issues with self closing tags without any values such as <abc//> or <StudentID//>

HyukjinKwon:
Yea, I think that's fixed at #303

Issue #398, opened 2019-08-02
Records dropped on large files when partition breaks rowTag


jimenefe:
We have found an issue where records are dropped silently.

By default, the input XML file will be partitioned into several "splits" with size as set in the 'fs.local.block.size' config setting (by default 32MB). We've found that when this split happens in the middle of the starting row tag element, the record is lost and not read into the DataFrame.

The easy workaround is to set 'fs.local.block.size' larger than the file size, which is guaranteed to work, but has a severe performance penalty for large files.

To check this scenario, I've setup a quick and dirty test as follows (inside `XmlSuite`):

```scala
test("fails on file partitioning") {

    val xml =
      s"""<people>
         |   <person>
         |     <age born="1990-02-24">25</age>
         |     <name>Hyukjin</name>
         |   </person>
         |   <person>
         |     <age born="1985-01-01">30</age>
         |     <name>Lars</name>
         |   </person>
         |   <person>
         |     <age born="1980-01-01">30</age>
         |     <name>Lion</name>
         |   </person>
         | </people>
      """.stripMargin

    val xmlFile = File.createTempFile("ages", ".xml")
    val writer = new BufferedWriter(new FileWriter(xmlFile))
    writer.write(xml)
    writer.close()

    println(s"Reading from $xmlFile")
    spark.stop() // stop the existing one setup in `beforeAll`
    spark = SparkSession.builder()
      .master("local[2]")
      .appName("XmlSuite")
      .config("spark.ui.enabled", false)
      .config("spark.hadoop.fs.local.block.size", "0x68")
      .getOrCreate()

    val results = spark.read
      .option("rowTag", "person")
      .option("mode", "FAILFAST")
      .format("xml")
      .load(xmlFile.getAbsolutePath)

    results.show(false)
  }
```

Setting block size to partition at "0x68" will break the second `<person>` element in the middle. Executing this code actually results in zero records being read... different symptom, but same underlying issue.

The problem seems to be inside `XmlInputFormat` where there are calls to `filePosition.getPos`. The goal seems to be to know the "index" of the character being read at any given time, so as to know when it should stop reading as it has gone beyond the split boundary

[...]

PR #399, opened 2019-08-02
Fix for data loss when input file partitioned through rowTag element

jimenefe:
Fixes issue reported in [390](https://github.com/databricks/spark-xml/issues/398)

### DIFF
--- a/src/main/scala/com/databricks/spark/xml/XmlInputFormat.scala
+++ b/src/main/scala/com/databricks/spark/xml/XmlInputFormat.scala
@@ -249,6 +249,11 @@ private[xml] class XmlRecordReader extends RecordReader[LongWritable, Text] {
         si = 0
         ei = 0
         depth -= 1
+      } else if (si == (startTag.length - 1) && Character.isWhitespace(c)) {
+        // found a start tag with attributes
+        si = 0
+        ei = 0
+        depth += 1
       } else {
         // Not in start tag or end tag.
         si = 0
--- a/src/main/scala/com/databricks/spark/xml/XmlInputFormat.scala
+++ b/src/main/scala/com/databricks/spark/xml/XmlInputFormat.scala
@@ -15,14 +15,14 @@
  */
 package com.databricks.spark.xml
 
-import java.io.{IOException, InputStream, InputStreamReader, Reader}
+import java.io.{ IOException, InputStream, InputStreamReader, Reader }
 import java.nio.charset.Charset
 
-import org.apache.hadoop.fs.Seekable
+import com.databricks.spark.xml.XmlRecordReader.PositionTracker
 import org.apache.hadoop.io.compress._
-import org.apache.hadoop.io.{LongWritable, Text}
-import org.apache.hadoop.mapreduce.{InputSplit, RecordReader, TaskAttemptContext}
-import org.apache.hadoop.mapreduce.lib.input.{FileSplit, TextInputFormat}
+import org.apache.hadoop.io.{ LongWritable, Text }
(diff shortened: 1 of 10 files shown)


src/main/scala/com/databricks/spark/xml/XmlInputFormat.scala
```diff
@@ -15,14 +15,14 @@
  */
 package com.databricks.spark.xml
 
-import java.io.{IOException, InputStream, InputStreamReader, Reader}
+import java.io.{ IOException, InputStream, InputStreamReader, Reader }
```
srowen: (Could we revert the spacing changes and reordering?)

src/main/scala/com/databricks/spark/xml/XmlInputFormat.scala
```diff
@@ -285,3 +283,18 @@ private[xml] class XmlRecordReader extends RecordReader[LongWritable, Text] {
...
+
+object XmlRecordReader {
+  trait PositionTracker extends Reader {
+    val initialPosition: Long
+
+    private[this] var offset: Long = 0
+
+    def getPosition: Long = initialPosition + offset
+
+    override def read(): Int = {
```
srowen: My concern here is that this isn't intercepting calls to `read(byte[])` etc. How about adding CountingInputStream from Common IO? it's a wrapper that's exactly meant to tell you how many bytes have been read through the input stream. I think that's what we need, added to the start position, to really know where in the file it has read.

srowen:
OK so I'm looking into this more today, and it's more complex than I thought.

First, there's a problem with tracking the number of _chars_ read, which is what this does. The end and start offsets are in bytes, and so 1 char is not always 1 byte. That's easy enough to address another way, like `CountingInputStream` I think (or wrapping the `InputStream` the `Reader` reads; we do in fact only rely on `.read()` now, but that could change).

I agree that using `FilePosition` is problematic because all paths here would end up with a buffered stream from Hadoop and so the read position may not match exactly how much has been consumed. However I returned to the source of `LineRecordReader` which this is based on, and it does use the same mechanism for checking how much has been read. Sometimes.

It really uses this only for compressed input. For uncompressed input this is easy, as `end` and `start` make it entirely possible to know when one has finished the split by counting bytes read. For compressed input, it's not clear how many uncompressed bytes come out of the source bytes between `start` and `end` so this won't work.

For unsplittable compression, this won't matter. One file is one split always. Reading until the stream gives EOF is sufficient.

For splittable compression, this again won't work, and I'm still trying to work out how it handles this case. You can see the source checks the 'adjusted' start and end values but a) I don't think this turns them into somehow 'uncompressed' offsets, and b) it doesn't look like they are set at all anyway by the one implementation, for bzip2.

I think we'll need more tests in the end to check what happens on .gz and .bzip2 input, but that's TBD.

srowen:
Bad news: `InputStreamReader` itself buffers from the underlying bytes. So it throws off any attempt to know exactly how many bytes have been processed.

It appears possible to hack this and get some internal `ByteBuffer` from it, and we can subtract off the number of bytes buffered but not processed yet. I have a version of this change that works, but it's hack. Yet I have no better ideas.

I'm pretty sure there are problems with compressed input too, and will work on some tests to check their behavior here.

srowen:
Check out https://github.com/databricks/spark-xml/pull/400

srowen:
We merged the other change, but credit to you for finding the issue and test case, and proposing a fix that was directionally correct.

jimenefe:
Thanks Sean, as long as the issue is fixed, I'm happy :-)

PR #400, opened 2019-08-04
Better handle rows that break across splits, and other small related fixes

srowen:
This attempt to address https://github.com/databricks/spark-xml/issues/398
See also https://github.com/databricks/spark-xml/pull/399

The change is I believe explained in comments below.


src/main/scala/com/databricks/spark/xml/XmlInputFormat.scala
```diff
@@ -106,11 +112,45 @@ private[xml] class XmlRecordReader extends RecordReader[LongWritable, Text] {
...
+      fsin.seek(start)
+      countingIn = new CountingInputStream(fsin)
+      in = countingIn
+      // don't use filePosition in this case. We have to count bytes read manually
     }
+
     reader = new InputStreamReader(in, charset)
+
+    if (codec == null) {
+      // Hack: in the uncompressed case (see more below), we must know how much the
```
HyukjinKwon: Yea ... I don't like this hack too ... but seems no better way.

srowen:
Thanks @HyukjinKwon - we may want to cut an 0.6.0 release for this, plus the `inferSchema` change. I may need to learn from you how to do it.

PR #412, opened 2019-10-28
Ignore processing instructions to fix errors caused by them

### LINKED ISSUE #411: Processing Instructions Cause Failed Import
When importing XML containing [processing instructions](https://www.w3.org/TR/xml/#dt-pi), spark-xml yields no data. Consider this test case:
```xml
<foo>
<bar>
<?issue?>
lorem ipsum
</bar>
</foo>
```
When using `sql.read.format("xml").option("rowTag", "foo").load(filepath)` in the Python API, the above yields an empty dataframe. In contrast, running the same code with the `<?issue?>` removed appears to create a valid dataframe containing a "bar" column with one entry, "lorem ipsum". 

For reference, my current submission command is `spark-submit --packages com.databricks:spark-csv_2.11:1.5.0,com.databricks:spark-xml_2.11:0.6.0 script.py`.

At the moment, there doesn't appear to be a way to handle this other than manually removing all the processing instructions. Is there some way to ignore them and access the rest of the contents?

srowen:
Simply ignore processing instructions, or else it will cause schema inference and parsing to fail when included inside an element.

Closes #411

### DIFF
--- a/src/main/scala/com/databricks/spark/xml/parsers/StaxXmlParser.scala
+++ b/src/main/scala/com/databricks/spark/xml/parsers/StaxXmlParser.scala
@@ -85,8 +85,11 @@ private[xml] object StaxXmlParser extends Serializable {
       factory.setProperty(XMLInputFactory.IS_COALESCING, true)
       val filter = new EventFilter {
         override def accept(event: XMLEvent): Boolean =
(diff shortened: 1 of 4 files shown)


HyukjinKwon:
Thanks @srowen.

Issue #415, opened 2019-10-30
XML not imported properly when tag has both value and attribute


gmoore016:
It seems as though certain tags fail to load when their siblings have text values as well as attributes. As an example:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<root>
<foo> issue <bar> lorem </bar> </foo>
<missing> ipsum </missing>
</root>
```
When I load this using `xml = sql.read.format("xml").option("rowTag", "root").option("inferSchema", "false").option("mode", "FAILFAST").load("example.xml")` then run `xml.printSchema()`, I get the following schema:
```
root
|-- bar: string (nullable = true)
|-- foo: string (nullable = true)
```
which is notably missing any reference to the `<missing>` tag from the example XML, and lists `<bar>` as a sibling of `<foo>` rather than as a child. However, I do not seem to have this problem when I remove `issue`. I double checked via [this website](https://www.w3schools.com/xml/xml_validator.asp), and the example appears to be valid XML. 
Is there any way I can access the text from `<missing>`?

srowen:
Hm, not sure. I know there are problems handling mixed text/element children well. The problem is it's not super clear how this maps to a tabular schema: foo is a string, with child bar that's a string? that's a fine interpretation in this case. But what if foo had more text children after bar? It does fine on less-ambiguous cases where the XML is basically a straight tabular data set encoded as XML.

That said I don't see a good reason it misses 'missing' here. What happen if you infer the schema though?

gmoore016:
When I run `xml = sql.read.format("xml").option("rowTag", "root").option("inferSchema", "true").load("example.xml")` and immediately run `xml.printSchema()` I receive the same schema as above.

To give a use case, I'm looking at text with in-text citations. For example, `<text> According to <ref> Turing (1936) </ref>, computable numbers are...</text>`. 

At least for this context, I might suggest an array containing each string. For example, given `<foo> lorem <bar> text </bar> ipsum </foo>`, you would have the `bar` element as well as the array `["lorem", "ipsum"]`. This would yield the schema:
```{xml}
root
|--foo
|    |--_values: array
|    |    |--element: string
|    |--bar: string
```
As an alternative, you may wish to simply store the full string; in the above case, you could have both the `bar` element as well as the string `"lorem <bar> text </bar> ipsum"`--though this may get quite large if the internal elements are complex. These both would make it relatively simple to construct the meaning of the original sentence while parsing the XML in a usable way.

srowen:
[...]

PR #416, opened 2019-11-03
Handle character elements mixed text/node elements

srowen:
(Don't merge this yet)

See https://github.com/databricks/spark-xml/issues/415 for a bug report. There are two issues:
1. We don't really handle mixed text / node children of elements, possibly by design
2. But, their presence in this example causes the rest of the normal XML to not be parsed correctly

We need to fix the second, I think, one way or the other. 

What about the first?

The conceptual issue is that an element may have character elements all over the place between its child nodes. What does that map to in a schema?
1. an array of strings under the same `_VALUE` element we use for the body of an element with attributes?
2. same, but one joined string?
3. same, but just the first string

Number 2 doesn't quite make sense as there's no conceptual reason to join them. Number 1 does, but, after hacking on this, looks like it's quite hard to implement. Number 3 was hackable; that's this PR. But I don't feel great about it, especially as it's a halfway design, to just take the first character element child and use that, and drop the others. (That would have, however, addressed the first issue in this particular #415 )

Another approach is to simply _correctly_ ignore all of that text. That is, just fix the second issue above.

That I think is also possible via a simplified version of this change.

I'm putting this out for conversation before going that way though.

### DIFF
--- a/src/main/scala/com/databricks/spark/xml/parsers/StaxXmlParser.scala
+++ b/src/main/scala/com/databricks/spark/xml/parsers/StaxXmlParser.scala
@@ -154,33 +154,23 @@ private[xml] object StaxXmlParser extends Serializable {
       case (c: Characters, ArrayType(st, _)) =>
         // For `ArrayType`, it needs to return the type of element. The values are merged later.
         convertTo(c.getData, st, options)
-      case (c: Characters, st: StructType) if st.fields.map(_.name).contains(options.valueTag) =>
+      case (c: Characters, st: StructType) =>
         // If a value tag is present, this can be an attribute-only element whose values is in that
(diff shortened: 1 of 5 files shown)


srowen:
I pushed a different change, to just ignore the text elements. This is simpler and perhaps more reliable, and while there's less output here, it may be less confusing to parse 0 rather than 1 of N character elements mixed in the other elements. It does address the parsing problem in #415 that prevented another unambiguous field from from being parsed.

srowen:
I'm going ahead with this because it's a narrow and unintrusive change, doesn't appear to change any desirable existing behavior, and allows the parser to proceed correctly where ignoring text children would yield a coherent output.

Issue #224, opened 2016-12-27
is there any way to refer XSD as a schema reference to process xml messages ?


skm235:

HyukjinKwon:
Currently, it does not have a functionality for it. Feel free to open a PR for this.

skm235:
I am sorry. Don't know how to do it.

Sokrates80:
that will be great actually.

MaciejGajewski:
I am also looking for this feature as I'm working with pretty big XMLs with plenty of optional tags. Would like to provide generic schema beforehand and have XML messages read into this schema (instead of infering the individual message schema) and have optional elements filled with nulls.

marshackVB:
Just curious if there is an update on this. If not, any tips on how to approach the problem posed above by MacieGajewski would be much appreciated.

srowen:
Someone would need to write a converter from XSD schema to a StructType for Spark. I'm not sure anyone is going to work on this, so it's open to those who want to submit a pull request for it.

akshaygehi:
**I was thinking of using a two step approach of solving this problem without code changes:**

- Use a library to generate from an XSD its sample XMLs including optional attributes. For example XSInstance allows you to do this. You could alternatively use online generators or desktop tools such as LiquidXML to do this. 

- Once you have the sample XML, create a dataframe from this and export the schema into a JSON file. 

Now you can use this exported structure file during the actual parsing i.e. with the actual XML files

srowen:
Maybe, as a general approach, though that seems kind of outside the scope of this library. Note that this can't work in the general case because XML and XSDs can follow schemas that don't conform to a tabular form.

akshaygehi:
I agree with you. It will be outside the scope of the library and will only work as a workaround. Can you elaborate what you mean by general case? Are you pointing at some limitation in Spark DataFrame which would restrict having complex structures or something else?

srowen:
No, simply that all DataFrames are tabular, and XML is not at all tabular.

developeratkiit:
I am also looking for such function where i need to transform the dataframe to xsd and then write it as  XML file.

srowen:
One thing I think we can do is simply validate the XML rows according to an XSD. It would still infer the schema separately, not from the XSD, but, presumably those are consistent if parsing succeeds.

developeratkiit:
Thanks for response.

I have got a requirement where i have to read the schema frok XSD file and
then convert the input DataFrame as per schema and write the xml
[...]

PR #419, opened 2019-11-17
Add setting to validate rows against an XSD

srowen:
This adds a new configuration `rowValidationXSDPath`, which gives a path to an XSD file. If specified, this is used to validate the XML of each row. If validation fails, the row is considered 'corrupt' and handled as usual.

This does not affect the schema inferred by or specified for XML parsing. It can be used for finer-grained validation beyond simple schema checking.

See https://github.com/databricks/spark-xml/issues/224 for some discussion.

One open question is: is a path to a file the best API? support general URIs?


srowen:
Oops, I realize there's a flaw here. The schema is read from local files, but is read from the executors. I need a different approach here, like reading it once and serializing it out to executors. Copying the files out with addFiles is possible, but, would require knowing all the dependent XSDs that are imported, which is hard.

srowen:
See https://github.com/databricks/spark-xml/pull/424 for an improvement and fix.

PR #420, opened 2019-11-18
Support Java 11 / test vs Spark 3

srowen:
This adds a test profile for Spark 3 and Java 11.
It switches to use the IndentingXMLStreamWriter from Glassfish instead of from the JDK, as it is no longer present in Java 9+.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 language: scala
 sudo: false
 jdk: openjdk8
+scala:
+  - 2.12.10
 cache:
   directories:
     - $HOME/.ivy2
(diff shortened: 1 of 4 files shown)


NOTICE
```diff
@@ -1,5 +1,46 @@
...
 Apache Commons IO
 Copyright 2002-2019 The Apache Software Foundation
 
 This product includes software developed at
-The Apache Software Foundation (http://www.apache.org/).
\ No newline at end of file
+The Apache Software Foundation (http://www.apache.org/).
+
+
+-----------------------------------
```
srowen: Wait a sec, I'm getting confused. We don't need these as we don't bundle anything in the distribution; we don't make any uber JAR here. Hm, I think I should pull these out.

srowen:
NB: I think the library already worked fine with Spark 3, from local testing. This just adds tests. This is really about JDK 9+ support

PR #421, opened 2019-11-18
Minor: NOTICE not needed as there is no uber JAR; bump version for MiMa purposes

srowen:


srowen:
I earlier added the NOTICE as is standard practice in Spark, but, totally forgot there is no uber JAR binary distribution here. This software is not redistributed by the project at all, so even following ASF-like best practices, not needed.

PR #423, opened 2019-12-04
Add xmlDataFrame convenience method for xmlRdd

srowen:
This simply adds a convenience method, xmlDataFrame, that accepts a DataFrame with one string col, instead of an RDD[String] as in xmlRdd.

### DIFF
--- a/src/main/scala/com/databricks/spark/xml/XmlReader.scala
+++ b/src/main/scala/com/databricks/spark/xml/XmlReader.scala
@@ -17,7 +17,7 @@ package com.databricks.spark.xml
 
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.{DataFrame, SparkSession, SQLContext}
-import org.apache.spark.sql.types.StructType
+import org.apache.spark.sql.types.{StringType, StructType}
 import com.databricks.spark.xml.util.XmlFile
 import com.databricks.spark.xml.util.FailFastMode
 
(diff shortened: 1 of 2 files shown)


src/main/scala/com/databricks/spark/xml/XmlReader.scala
```diff
@@ -112,6 +117,23 @@ class XmlReader extends Serializable {
     spark.baseRelationToDataFrame(relation)
   }
 
+  /**
+   * @param spark current SparkSession
+   * @param df XML for individual 'rows' as Strings
+   * @return XML parsed as a DataFrame
+   */
+  def xmlDataFrame(spark: SparkSession, df: DataFrame): DataFrame = {
```
HyukjinKwon: Oh, wait, how about changing `DataFrame` -> `Dataset[String]`?

srowen: I figured the common case might be parsing the result of `myDF.select("someXML")`.
Want to add xmlDataset as well for Dataset[String]? it's trivial.

HyukjinKwon: I just wanted to make it matched to existing Spark API like https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/DataFrameReader.scala#L506

Actually I have an idea about adding an implicit too so that our library allows `spark.read.xml(dataset)`. Let me push some changes into this branch.

HyukjinKwon: Actually, let me just merge and make a separate PR. Can you resolve conflicts?

HyukjinKwon:
Actually I resolved it and merged it by myself at https://github.com/databricks/spark-xml/commit/935bc9693b7b0431b032ddd5a74df52695ea7870 :D.

src/main/scala/com/databricks/spark/xml/XmlReader.scala
srowen: Yeah go for it, any helper or convenience methods here are probably worthwhile.

PR #424, opened 2019-12-04
Improve XSD validation: support files added via addFile(); cache Schema

srowen:
(@HyukjinKwon only if you care to review, no big deal)

I found after adding the XSD validation feature that creating the XSD schema can be expensive. It can refer to lots of external resources that take time to download, and are accessed repeatedly. Further, I realized that the existing solution won't work unless the XSD itself is available on executors, unless the XSD is already on an accessible distributed store.

I'd propose to cache the Schema, firstly, to avoid recomputing it for inference vs parsing and for every partition. I've also made it support files added by SparkContext.addFile

### DIFF
--- a/src/main/scala/com/databricks/spark/xml/util/ValidatorUtil.scala
+++ b/src/main/scala/com/databricks/spark/xml/util/ValidatorUtil.scala
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2019 Databricks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


HyukjinKwon:
ah .. I missed it too ...

src/main/scala/com/databricks/spark/xml/util/ValidatorUtil.scala
```diff
@@ -0,0 +1,55 @@
...
+ */
+private[xml] object ValidatorUtil {
+
+  // Parsing XSDs may be slow, so cache them by path:
+
+  private val cache = CacheBuilder.newBuilder().softValues().build(
+    new CacheLoader[String, Schema] {
+      override def load(key: String): Schema = {
+        // Handle case where file exists as specified
+        var path = Paths.get(key)
```
HyukjinKwon: Maybe we could use Hadoop's Path to allow to access on HDFS too, but I think it can be done separately as well.

srowen: It's a good point. I had in mind that the common case would be a local file but, hm, maybe not. Hadoop's API would support local files too right?

HyukjinKwon: Yeah, it can read local files too. Can be done separately I think. Let me just merge for now.

PR #429, opened 2019-12-27
from_xml and schema_for_xml support

srowen:
This is a continuation of https://github.com/databricks/spark-xml/pull/334/

This implements a `from_xml` expression that can turn an XML string column into a parsed structured column. It also opens up `inferSchema` for a `Dataset[String]` as a necessary support function.

The rest is really mild refactoring.

### DIFF
--- a/src/main/scala/com/databricks/spark/xml/XmlDataToCatalyst.scala
+++ b/src/main/scala/com/databricks/spark/xml/XmlDataToCatalyst.scala
@@ -35,7 +35,7 @@ case class XmlDataToCatalyst(
 
   @transient
   private lazy val factory: XMLInputFactory = StaxXmlParserUtils.buildFactory()
-  
+
   override lazy val dataType: DataType = schema
 
   override def checkInputDataTypes(): TypeCheckResult = schema match {
--- a/src/main/scala/com/databricks/spark/xml/XmlDataToCatalyst.scala
+++ b/src/main/scala/com/databricks/spark/xml/XmlDataToCatalyst.scala
@@ -1,0 +1,58 @@
+/*
+ * Copyright 2019 Databricks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.databricks.spark.xml
+
+import javax.xml.stream.XMLInputFactory
+
+import org.apache.spark.sql.catalyst.CatalystTypeConverters
+import org.apache.spark.sql.catalyst.analysis.TypeCheckResult
+import org.apache.spark.sql.catalyst.expressions.codegen.CodegenFallback
+import org.apache.spark.sql.catalyst.expressions.{ExpectsInputTypes, Expression, UnaryExpression}
+import org.apache.spark.sql.types._
+import org.apache.spark.unsafe.types.UTF8String
+
+import com.databricks.spark.xml.parsers.{StaxXmlParser, StaxXmlParserUtils}
+
+case class XmlDataToCatalyst(
+    child: Expression,
+    schema: StructType,
+    options: XmlOptions)
+  extends UnaryExpression with CodegenFallback with ExpectsInputTypes {
+
+  @transient
+  private lazy val factory: XMLInputFactory = StaxXmlParserUtils.buildFactory()
+  
+  override lazy val dataType: DataType = schema
+
+  override def checkInputDataTypes(): TypeCheckResult = schema match {
+    case _: StructType =>
+      super.checkInputDataTypes()
+    case _ => TypeCheckResult.TypeCheckFailure(
+      s"Input schema ${schema.simpleString} must be a struct or an array of structs.")
+  }
+
+  override def nullSafeEval(xml: Any): Any = xml match {
+    case string: UTF8String =>
+      CatalystTypeConverters.convertToCatalyst(
+        StaxXmlParser.parseColumn(string.toString, schema, factory, options))
+    case string: String =>
+      StaxXmlParser.parseColumn(string.toString, schema, factory, options)
+    case _ => null
+  }
+
+  override def inputTypes: Seq[DataType] = StringType :: Nil
+}
--- a/src/main/scala/com/databricks/spark/xml/XmlDataToCatalyst.scala
+++ b/src/main/scala/com/databricks/spark/xml/XmlDataToCatalyst.scala
@@ -17,9 +17,9 @@
 package com.databricks.spark.xml
 
 import org.apache.spark.sql.catalyst.CatalystTypeConverters
-import org.apache.spark.sql.catalyst.analysis.TypeCheckResult
 import org.apache.spark.sql.catalyst.expressions.codegen.CodegenFallback
 import org.apache.spark.sql.catalyst.expressions.{ExpectsInputTypes, Expression, UnaryExpression}
+import org.apache.spark.sql.catalyst.util.GenericArrayData
 import org.apache.spark.sql.types._
 import org.apache.spark.unsafe.types.UTF8String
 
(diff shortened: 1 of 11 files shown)


srowen:
CC @SpaceRangerWes and @HyukjinKwon for comment.
The only thing I don't quite like is that you have to separately infer the schema for the column, then apply it, but, otherwise pretty fine IMHO.

srowen:
(Also I'd need to update docs accordingly to describe the new functionality)

src/main/scala/com/databricks/spark/xml/package.scala
```diff
@@ -19,10 +19,64 @@ import scala.collection.Map
...
+import org.apache.spark.annotation.Experimental
 import org.apache.spark.sql._
-import com.databricks.spark.xml.util.XmlFile
+import org.apache.spark.sql.catalyst.expressions.Expression
+import org.apache.spark.sql.catalyst.parser.CatalystSqlParser
+import org.apache.spark.sql.types.StructType
+
+import com.databricks.spark.xml.util.{InferSchema, XmlFile}
 
 package object xml {
```
HyukjinKwon: Maybe we should create a package called `functions.scala` under this `xml` directory to match it to `avro`'s functions or other functions in Apache Spark.

src/main/scala/com/databricks/spark/xml/package.scala
```diff
@@ -19,10 +19,64 @@ import scala.collection.Map
...
+    InferSchema.infer(ds.rdd, XmlOptions(options.toMap))
+
+  /**
+   * Parses a column containing a XML string into a `StructType` with the specified schema.
+   *
+   * @param e a string column containing XML data
+   * @param schema the schema to use when parsing the XML string
+   */
+  @Experimental
+  implicit def from_xml(e: Column, schema: StructType): Column =
```
HyukjinKwon: If https://github.com/databricks/spark-xml/pull/429/files#r361887418 is addressed, I am sure it doesn't need to be an implicit function.

src/main/scala/com/databricks/spark/xml/package.scala
```diff
@@ -19,10 +19,64 @@ import scala.collection.Map
...
-import com.databricks.spark.xml.util.XmlFile
+import org.apache.spark.sql.catalyst.expressions.Expression
+import org.apache.spark.sql.catalyst.parser.CatalystSqlParser
+import org.apache.spark.sql.types.StructType
+
+import com.databricks.spark.xml.util.{InferSchema, XmlFile}
 
 package object xml {
+
+  private def withExpr(expr: Expression): Column = new Column(expr)
```
HyukjinKwon: Seems we don't need this very much ...

src/main/scala/com/databricks/spark/xml/package.scala
```diff
@@ -19,10 +19,64 @@ import scala.collection.Map
...
+
+  /**
+   * Infers the schema of XML documents as strings.
+   *
+   * @param ds Dataset of XML strings
+   * @param options additional XML parsing options
+   * @return inferred schema for XML
+   */
+  @Experimental
+  def inferSchema(ds: Dataset[String], options: Map[String, String]): StructType =
```
HyukjinKwon: There's `schema_of_json` and `schema_of_csv` in Apache spark. We can don't add both in this PR I believe. I can work on both separately.

src/main/scala/com/databricks/spark/xml/package.scala
```diff
@@ -19,10 +19,64 @@ import scala.collection.Map
...
+   * Parses a column containing a XML string into a `StructType` with the specified schema.
+   *
+   * @param e a string column containing XML data
+   * @param schema the schema to use when parsing the XML string
+   * @param options key-value pairs that correspond to those supported by [[XmlOptions]]
+   */
+  @Experimental
+  implicit def from_xml(e: Column, schema: StructType, options: Map[String, String]): Column =
+    withExpr {
+      val map = options + ("isFunction" -> "true")
```
HyukjinKwon: seems we don't need `isFunction`

src/main/scala/com/databricks/spark/xml/package.scala
srowen: Do you mean this should be called `schema_of_xml`, at least to start?
I think we have to have such a method, or there's no way to get the schema to pass to `from_xml`.

srowen:
@HyukjinKwon thank you - I believe I addressed all your previous comments, and new ones. The only item seems to be what to do about inferring schema. Do you have better ideas there?

src/main/scala/com/databricks/spark/xml/package.scala
HyukjinKwon: Yeah, I meant we should at least name it `schema_of_xml`.

Also, there have been many discussions about implementation details with multiple Spark committers ...

it ended up with implementing an expression that infers the schema from single record (single JSON or CSV), instead of going through all data.

If you don't mind, I will work on that. I should explain the history in the PR description I suspect.

src/main/scala/com/databricks/spark/xml/functions.scala
```diff
@@ -0,0 +1,75 @@
...
+import org.apache.spark.sql.{Column, Dataset}
+import org.apache.spark.sql.catalyst.parser.CatalystSqlParser
+import org.apache.spark.sql.types.StructType
+
+import com.databricks.spark.xml.util.InferSchema
+
+/**
+ * Support functions for working with XML columns directly.
+ */
+object functions {
```
HyukjinKwon: Sean, is the style rule change because of this? I think we could just ignore this for now (as Spark side does):

```scala
// scalastyle:off
object functions {
// scalastyle:on
```

srowen: I figured, this is the only object the style rule covers, so, if it doesn't match the rule, the rule can be relaxed? it really won't matter either way

srowen:
Some possible TODOs, which I might pursue in another PR:

- A `from_xml` function that operates on a String, for direct use in a UDF
- Expand the support (as in the original PR) to operate on an array of strings, and return an array of structs, if possible. That helps if the XML is in an array of strings

src/main/scala/com/databricks/spark/xml/XmlDataToCatalyst.scala
```diff
@@ -0,0 +1,60 @@
...
+    case ArrayType(st: StructType, _) => st
+  }
+
+  override def nullSafeEval(xml: Any): Any = xml match {
+    case string: UTF8String =>
+      CatalystTypeConverters.convertToCatalyst(
+        StaxXmlParser.parseColumn(string.toString, rowSchema, options))
+    case string: String =>
+      StaxXmlParser.parseColumn(string.toString, rowSchema, options)
+    case arr: GenericArrayData =>
```
srowen: @HyukjinKwon not sure if you want to take another look, but here I tried to add support for arrays, which will be handy in some cases, so you can parse an array of strings to an array of structs directly.

src/main/scala/com/databricks/spark/xml/functions.scala
```diff
@@ -0,0 +1,81 @@
...
+
+  /**
+   * Infers the schema of XML documents when inputs are arrays of strings, each an XML doc.
+   *
+   * @param ds Dataset of XML strings
+   * @param options additional XML parsing options
+   * @return inferred schema for XML. Will be an ArrayType[StructType].
+   */
+  @Experimental
+  def schema_of_xml_array(ds: Dataset[Array[String]],
```
srowen: Also, new method for getting the ArrayType schema out of an Array of strings. Not super pretty but helpful

src/main/scala/com/databricks/spark/xml/functions.scala
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  /**
+   * @param xml XML document to parse, as string
+   * @param schema the schema to use when parsing the XML string
+   * @param options key-value pairs that correspond to those supported by [[XmlOptions]]
+   * @return [[Row]] representing the parsed XML structure
+   */
+  @Experimental
+  def from_xml_string(xml: String, schema: StructType,
```
srowen: One new method for just directly going from String to Row, for UDF usage

src/main/scala/com/databricks/spark/xml/functions.scala
HyukjinKwon: Hm, @srowen, I think this can be elegantly worked around via:

```scala
df.select(transform(col("array_of_xml"), x => from_xml(x)))
``` 

Although those were added as of Spark 3 (SPARK-27297).

e.g.)

```scala
scala> spark.range(1).selectExpr("array(1,2,3) as arr").select(transform(col("arr"), x => concat(x))).printSchema()
root
 |-- transform(arr, lambdafunction(concat(x), x)): array (nullable = false)
 |    |-- element: string (containsNull = false)
```

src/main/scala/com/databricks/spark/xml/XmlDataToCatalyst.scala
HyukjinKwon: I think it can be covered by existing APIs (https://github.com/databricks/spark-xml/pull/429/files#r363134329). I think it's a bit awkward to mix Spark type and XML strings ...

src/main/scala/com/databricks/spark/xml/functions.scala
```diff
@@ -0,0 +1,81 @@
...
+
+  /**
+   * Infers the schema of XML documents as strings.
+   *
+   * @param ds Dataset of XML strings
+   * @param options additional XML parsing options
+   * @return inferred schema for XML
+   */
+  @Experimental
+  def schema_of_xml(ds: Dataset[String], options: Map[String, String] = Map.empty): StructType =
```
HyukjinKwon: Sorry for a bit of forth and back. I believe `snake_case` naming was introduced fro compatibility of SQL as of SPARK-10621. I think `schema_of_xml` and `from_xml_string` are actually not the SQL-like functions as it takes whole datasets or take other types of signature.

Maybe it's better to move both to `https://github.com/databricks/spark-xml/blob/master/src/main/scala/com/databricks/spark/xml/package.scala` as proposed before just, as static functions (or as an implicit against `Dataset[String]` if applicable).

src/main/scala/com/databricks/spark/xml/functions.scala
srowen: Hm, so can this work in Spark 2.x then? we'd still have to support it.
I agree that's much more elegant for Spark 3 though

src/main/scala/com/databricks/spark/xml/functions.scala
srowen: So move all of functions.scala to package.scala? and rename to something like inferSchema again?

HyukjinKwon: I meant to move, from `functions.scala` to `package.scala` with renaming, only functions that don't have `Column` signature because the functions with `Column` signature as input/output are considered as SQL functions. Namely, we might have to move `schema_of_xml`, `schema_of_xml_array` and `from_xml_string`.

src/main/scala/com/databricks/spark/xml/functions.scala
HyukjinKwon: Okay .. I think we can deprecate this behaviour later when Spark XML starts to deprecate Spark 2 supports.

srowen:
@HyukjinKwon I moved the methods, but didn't rename. If you like, have another look.

HyukjinKwon:
I'll just merge it in and make a followup later if there;s something I find that I want to fix.

srowen:
OK. I was going to release 0.8.0 maybe next week, unless you spot something critical that shouldn't be left for 0.9.0.

HyukjinKwon:
I don't think there would be anything critical so should be fine.

PR #439, opened 2020-02-17
extract to a method the inference of the xml schema

caenrique:
This change is intended to make available the functionality of inferring the schema of a xml String.
This allows to have a schema inference over a RDD row by row, using a UDF to map over the desired field. This field being a String containing the whole xml document.

In my case, I needed this functionality to check if the inferred schema from the data is a subset of the schema that I want to impose on it. Every file can have problems, so loading a whole directory of xml files into a DataFrame directly wouldn't allow me to discard the bad files.

### DIFF
--- a/src/main/scala/com/databricks/spark/xml/package.scala
+++ b/src/main/scala/com/databricks/spark/xml/package.scala
@@ -115,6 +115,17 @@ package object xml {
   }
 
   /**
+   * Infers the schema of an XML document as a string.
+   *
+   * @param s XML string
+   * @param options additional XML parsing options
+   * @return inferred schema for XML
+   */
+  @Experimental
+  def schema_of_xml_from_string(s: String, options: Map[String, String] = Map.empty): StructType =
+    InferSchema.infer(s, XmlOptions(options))
+
+  /**
    * Infers the schema of XML documents as strings.
    *
    * @param ds Dataset of XML strings
--- a/src/main/scala/com/databricks/spark/xml/util/InferSchema.scala
+++ b/src/main/scala/com/databricks/spark/xml/util/InferSchema.scala
@@ -34,7 +34,7 @@ import com.databricks.spark.xml.util.TypeCast._
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.types._
 
-private[xml] object InferSchema {
+object InferSchema {
 
   /**
    * Copied from internal Spark api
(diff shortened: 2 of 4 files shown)


srowen:
Hm, but you wouldn't want to infer off one example in general, right? each row could appear to have a different schema depending on what fields are present.
If you know the schema you want, you just parse with it. It can kick out the bad records into a separate column, or just drop them.
You can also use an XSD for this.

caenrique:
Yes, I'm using other functions to check whether the inferred schema will fit my final schema or not. 
Parsing directly with the schema I want is not an option since the way spark works it will raise a NullPointerException when actually accessing the data if some non nullable field is missing, and It is a possibility for me to have malformed data.
The xsd definition of my data is a mess and also doesn't work for me in some cases because is too general (with some fields groups being a choice between varios fields and mandatory at the same time)

srowen:
In that case, it should fail to read the record and it's handled according to PERMISSIVE, FAILFAST, etc modes. What do you want it to do in this case? if you want to see the record, PERMISSIVE should make this NPE cause it to put it in the bad records column. If that isn't what happens that is a bug.

caenrique:
I want it to pase the good records and log the file name of the bad ones. For what I could see with the modes, it only works if the xml itself is malformed or of you provide the xsd (is that the case?), but providing the StructType schema it doesn't fail when parsing, even though there are missing fields. So later on, the DataFrameWritter will fail.
If this behavior is a bug, then I guess I can prepare a small example to reproduce it

srowen:
I see. The file name isn't generally available here.
It should handle the record as bad no matter why, including an exception during parsing.
Missing optional fields aren't an error. I'd have to see an example to figure out what you're seeing.

You can use `from_xml` and your schema to apply the parsing to a DataFrame of, say, file name and file contents. You apply it as a UDF to the contents. The struct you get back will either be the contents or contain a bad records member. You could then select rows with file name where the parse result indicates a bad record.

caenrique:
Yes, I'm actually doing exactly that but I needed this to implement a filter first. Missing optional fields is ok, missing mandatory fields is the problem.

I will prepare a example tomorrow to make it more clear. Thanks

caenrique:
I've added a couple of test cases to show the purpose of this change. Both the number of files and the XML data itself are obviously simplified. In a case with much complex nested structures the function `isSubset` is also used as a way to generate a `Validated[NonEmptyChain]` with all the differences between the two `StructTypes` so it helps a lot for debugging

srowen:
I'm still not sure why you'd infer a different schema for each message. You want one schema, you just want to customize parsing right?

caenrique:
yes, I don't want a different schema for each message, but I need to know the structure of each message to compare it with the expected schema and discard that message while being able to generate a diff

srowen:
OK, then I think you want to define the schema for parsing upfront. You can do that manually, or by inferring a schema and modifying it. I do not think we need methods to infer a schema on individual XML elements.

You can use PERMISSIVE mode to kick out any records that don't parse correctly to a new column. From there you can do whatever additional processing you want.

Issue #446, opened 2020-04-28
ClassCastException while reading an xml


GeethanadhP:
I am trying to process xml's from kafka using the `from_xml` which is throwing ClassCastException
After a day of debugging to find the xml and tag causing it, found that a particular tag is having spaces/tabs (or looks like spaces) for a `double` field

But my expectation is it should have gone into `_corrupt_record` when using PERMISSIVE mode or got dropped when using DROPMALFORMED, but neither is happening

**Note:** I wrote it to a file and try to read from the file also (just to check if its because of `from_xml`), but that failed too

Below is the full stacktrace

```
org.apache.spark.SparkException: Job aborted due to stage failure: Task 0 in stage 1.0 failed 4 times, most recent failure: Lost task 0.3 in stage 1.0 (TID 7, 10.150.51.6, executor 1): java.lang.ClassCastException: org.apache.spark.unsafe.types.UTF8String cannot be cast to java.lang.Double
	at scala.runtime.BoxesRunTime.unboxToDouble(BoxesRunTime.java:114)
	at org.apache.spark.sql.catalyst.expressions.BaseGenericInternalRow$class.getDouble(rows.scala:44)
	at org.apache.spark.sql.catalyst.expressions.GenericInternalRow.getDouble(rows.scala:195)
	at org.apache.spark.sql.catalyst.expressions.GeneratedClass$SpecificUnsafeProjection.writeFields_1_0$(Unknown Source)
	at org.apache.spark.sql.catalyst.expressions.GeneratedClass$SpecificUnsafeProjection.writeFields_4_0$(Unknown Source)
	at org.apache.spark.sql.catalyst.expressions.GeneratedClass$SpecificUnsafeProjection.writeFields_38_1$(Unknown Source)
	at org.apache.spark.sql.catalyst.expressions.GeneratedClass$SpecificUnsafeProjection.writeFields_41_0$(Unknown Source)
	at org.apache.spark.sql.catalyst.expressions.GeneratedClass$SpecificUnsafeProjection.apply(Unknown Source)
	at org.apache.spark.sql.catalyst.expressions.GeneratedClass$SpecificUnsafeProjection.apply(Unknown Source)
	at scala.collection.Iterator$$anon$11.next(Iterator.scala:410)
	at org.apache.spark.sql.execution.collect.UnsafeRowBatchUtils$.encodeUnsafeRows(UnsafeRowBatchUtils.scala:63)
[...]

PR #447, opened 2020-04-29
Handle case of whitespace-only, numeric-primitive field correctly

srowen:
Where a field has, say, type 'double', and the element value is just whitespace, it fails with a hard error instead of being a 'normal' parsing failure. See https://github.com/databricks/spark-xml/issues/446

This fixes it by actually trying to parse the whitespace as the desired type, to generate the normal failure. Otherwise the whitespace was returned to Spark as if it were a Double.

### DIFF
--- a/src/main/scala/com/databricks/spark/xml/parsers/StaxXmlParser.scala
+++ b/src/main/scala/com/databricks/spark/xml/parsers/StaxXmlParser.scala
@@ -153,19 +153,6 @@ private[xml] object StaxXmlParser extends Serializable {
           ""
         }
       case (_: EndElement, _: DataType) => null
-      case (c: Characters, _: DataType) if c.isWhiteSpace =>
-        // When `Characters` is found, we need to look further to decide
-        // if this is really data or space between other elements.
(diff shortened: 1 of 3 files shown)


src/main/scala/com/databricks/spark/xml/parsers/StaxXmlParser.scala
```diff
@@ -187,6 +174,18 @@ private[xml] object StaxXmlParser extends Serializable {
...
+      case (c: Characters, _: DataType) if c.isWhiteSpace =>
+        // When `Characters` is found, we need to look further to decide
+        // if this is really data or space between other elements.
+        val data = c.getData
+        parser.next
+        parser.peek match {
+          case _: StartElement => convertComplicatedType(dataType)
+          case _: EndElement if data.isEmpty => null
+          case _: EndElement if options.treatEmptyValuesAsNulls => null
+          case _: EndElement => convertTo(data, dataType, options)
```
srowen, line 186: This line changed; otherwise, the block just moved.

PR #448, opened 2020-05-25
fix date and datetime parsing

seddonm1:
Hi,
Currently the parsing of Date and Timestamp types does not comply with the the XSD schema definition: https://www.w3.org/TR/xmlschema-2/#dateTime

This PR adds the official date and datetime parsing when using a user-defined schema.

### DIFF
--- a/src/main/scala/com/databricks/spark/xml/util/TypeCast.scala
+++ b/src/main/scala/com/databricks/spark/xml/util/TypeCast.scala
@@ -83,7 +83,8 @@ object TypeCast {
   }
 
   val supportedXmlDateFormatters = List(
-    // '2011-12-03+01:00'; '2011-12-03'
+    // 2011-12-03
+    // 2011-12-03+01:00
     DateTimeFormatter.ISO_DATE
   )
 
@@ -102,7 +103,7 @@ object TypeCast {
   }
 
   val supportedXmlTimestampFormatters = List(
-    // "2002-05-30 21:46:54"
+    // 2002-05-30 21:46:54
     new DateTimeFormatterBuilder()
       .parseCaseInsensitive()
       .append(DateTimeFormatter.ISO_LOCAL_DATE)
--- a/src/main/scala/com/databricks/spark/xml/util/TypeCast.scala
+++ b/src/main/scala/com/databricks/spark/xml/util/TypeCast.scala
@@ -18,9 +18,9 @@ package com.databricks.spark.xml.util
 import java.math.BigDecimal
 import java.sql.{Date, Timestamp}
 import java.text.NumberFormat
-import java.util.Locale
-import java.time.format.DateTimeFormatter
 import java.time.{LocalDate, ZoneId, ZonedDateTime}
+import java.time.format.{DateTimeFormatter, DateTimeFormatterBuilder}
+import java.util.Locale
 
 import scala.util.Try
 import scala.util.control.Exception._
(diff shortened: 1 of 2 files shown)


src/main/scala/com/databricks/spark/xml/util/TypeCast.scala
```diff
@@ -80,6 +88,34 @@ object TypeCast {
...
+      case Nil => throw new IllegalArgumentException(s"cannot convert value $value to Date")
+      case head :: tail =>
+        try {
+          java.sql.Date.valueOf(LocalDate.parse(value, head))
+        } catch {
+          case e: Exception =>
+            parseXmlDate(tail, value)
+        }
+    }
+  }
```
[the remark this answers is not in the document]
seddonm1: sorry i usually check that but missed it. fixed.

seddonm1:
> Yeah it makes some sense. The semantics here aren't necessarily that of an XSD, but that's a reasonable standard to follow. 

This is actually the first PR for better handling of XML data with XSD provided. I have code that needs some cleanup that will generate the correct Spark StructType from the XSD.

> Right now I guess the problem is mostly that it's timezone-dependent when no timezone is specified?

Yes, this seems to be an oversight in the standard as you can't represent an Instant without a timezone. My view it is better to have a deterministic outcome (UTC) when the user has not provided the timezone information then at least date math can be used to override if needed.

> Really it would be better to use Spark's time/date handling code here, like in DateTimeUtils, but it might not be fully exposed.

Maybe but for the complexity of the code involved I am not sure it is beneficial.

> It's a behavior change but probably worth making to make this a little more robust.

To minimise impact the previous `Timestamp.valueOf` and `Date.valueOf` behaviors could be appended to the valid formatter list?

src/main/scala/com/databricks/spark/xml/util/TypeCast.scala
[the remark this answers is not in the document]
srowen: There are more now in the test code: https://travis-ci.org/github/databricks/spark-xml/jobs/691073458. I think you can run sbt checkstyle to check it locally if you like.

src/main/scala/com/databricks/spark/xml/util/TypeCast.scala
```diff
@@ -80,6 +88,34 @@ object TypeCast {
     }
   }
 
+  @scala.annotation.tailrec
+  def parseXmlDate(formatters: List[DateTimeFormatter], value: String): Date = {
```
srowen, line 92: I don't mind this construction with `tailrec`; I usually just write

```
for (formatter <- formatters) {
  try {
    return ...valueOf...
  } catch {
    _: Exception => // continue
  }
}
throw new IllegalArgumentException(...)
```

seddonm1:
> To confirm this should continue to work as before when a timezone is supplied right? will just default to UTC when not specified? that's good.

Correct. You can see examples in the tests.

> Are there any inputs that fail to parse now that would parse before, that you know of? I'd favor keeping as much behavior as possible without supporting the ambiguous local timezone-dependent parsing of course.

I have re-added support for `2002-05-30 21:46:54` style dates. These are almost [ISO_LOCAL_DATE_TIME](https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html#ISO_LOCAL_DATE_TIME) format but missing the `T` separator. This required a minor refactor to prevent creating the formatter objects each value.

src/main/scala/com/databricks/spark/xml/util/TypeCast.scala
seddonm1: This is a personal choice and as you are the maintainer I am happy to change to your style if you would like?

srowen: Eh, I don't feel strongly about it. Won't matter. I find the for loop simpler but I get the Scala construct here

srowen:
Again just FYI @HyukjinKwon , no action needed unless you spot anything that jumps out as problematic here.

Issue #449, opened 2020-05-25
PROPOSAL: Spark Schema generation from XSD Schema


seddonm1:
Now that `spark-xml` supports [XML Schema](https://www.w3.org/XML/Schema) definitions (XSD files) it would be good to use the schema information to generate the Spark Dataframe schema instead of either relying on schema-inference or having to manually create the schema.

To do this I think we need a few things:

- agreement that this is an idea worth pursuing.
- agreement that it is ok to add an import of a library like the [Apache Web Services XmlSchema](https://ws.apache.org/xmlschema/) to parse the XSD.
- sufficient tests.

I have already build a proof of concept which correctly parses the XSD to a StructType using the `XmlSchema` library above (which identified issues with the DateTime and Date types https://github.com/databricks/spark-xml/pull/448). I have tested this against a set of [ISO20022](https://www.iso20022.org/iso-20022-message-definitions) messages.

My challenge is I only have a limited knowledge of XSD so having correct XSD and valid XML files to run against and some help on this would be very useful. I am not sure how representative the `ISO20022` XSDs really are.

srowen:
Yeah I think that could be worthwhile. Not all XSDs define a schema that is tabular, so it would necessarily only work on certain simple types. If it covers enough ground to be nontrivial then I think it could be worthwhile even if understood to be experimental and not 100% going to work with all XSDs.

seddonm1:
Great. The current approach already supports deeply nested and complexTypes. See: https://www.iso20022.org/message/20821/download for an example.

I will prepare a PR and maybe we can work through appropriate tests together?

smoshah:
Hi, we are currently in a situation where we will have to define a huge XSD equivialant schema in spark to have a consistent ETL pipeline from our XML based source. If we had this option, it would save us days. It is definitely worth pursing.

seddonm1:
Hi,
I have put a gist here: https://gist.github.com/seddonm1/6e4354f68d7d74f71bb5e27d31d5b980

It needs work as I do not have many XSD schemas to play with but it seems to process https://www.iso20022.org/iso-20022-message-definitions schemas correctly.

@smoshah please let me know if it works and we can try to get it moving.

srowen:
Yeah if it does something useful, even if incomplete, we can add it and mark as 'experimental'. Open a PR when ready.

smoshah:
Hi,
[...]

PR #457, opened 2020-06-16
XSD -> schema tool with a test

srowen:
Relates to #449 and @seddonm1 's gist.

Adds a basic XSDToSchema utility that can parse a Spark schema from _some_ XSDs, those defining a table-like schema with simple, complex and sequence types.

### DIFF
--- a/src/main/scala/com/databricks/spark/xml/util/XSDToSchema.scala
+++ b/src/main/scala/com/databricks/spark/xml/util/XSDToSchema.scala
@@ -70,8 +70,7 @@ object XSDToSchema {
     schemaType match {
       // xs:simpleType
       case simpleType: XmlSchemaSimpleType =>
-        val baseName = "baseName"
-        simpleType.getContent match {
+        val schemaType = simpleType.getContent match {
           case restriction: XmlSchemaSimpleTypeRestriction =>
             val matchType =
               if (restriction.getBaseTypeName == Constants.XSD_ANYSIMPLETYPE) {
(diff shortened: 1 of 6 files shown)


src/main/scala/com/databricks/spark/xml/util/XSDToSchema.scala
```diff
@@ -0,0 +1,195 @@
...
+@Experimental
+object XSDToSchema {
+
+  /**
+   * Reads a schema from an XSD file.
+   *
+   * @param xsdFile XSD file
+   * @return Spark-compatible schema
+   */
+  @Experimental
```
srowen, line 40: I added these experimental "API" methods

src/main/scala/com/databricks/spark/xml/util/XSDToSchema.scala
```diff
@@ -0,0 +1,195 @@
...
+                   Constants.XSD_UNSIGNEDLONG |
+                   Constants.XSD_UNSIGNEDSHORT =>
+                StructField(baseName, IntegerType)
+              case Constants.XSD_LONG => StructField(baseName, LongType)
+              case Constants.XSD_STRING |
+                   Constants.XSD_BASE64 |
+                   Constants.XSD_ANYTYPE =>
+                StructField(baseName, StringType)
+            }
+          case _ => StructField(baseName, StringType)
```
srowen, line 116: This is about the only substantive change, which helps with 'any' type fields with no further content restrictions. Just treat them as strings. Other changes to the code are cosmetic simplifications (IMHO)

seddonm1: Agree.

src/main/scala/com/databricks/spark/xml/util/XSDToSchema.scala
```diff
@@ -80,41 +79,35 @@ object XSDToSchema {
                 restriction.getBaseTypeName
               }
             matchType match {
-              case Constants.XSD_BOOLEAN => StructField(baseName, BooleanType)
-              case Constants.XSD_BYTE => StructField(baseName, BinaryType)
-              case Constants.XSD_DATE |
```
seddonm1: Looks like you removed Date/DaateTime? You could release this with https://github.com/databricks/spark-xml/pull/448

srowen: I just let them fall into the general string case. But yeah maybe just as well to try them as a date or time type.

vim89:
Can I pull this version & compile myself? Before 0.10 gets released

srowen:
Yes you can just build it with SBT and try it. I should just make an 0.10 release soon.

PR #462, opened 2020-07-09
Rewrite of Schema inference, and small improvements

jimenefe:
* Root element with text and sub elements no longer discard the text
* Ability to treat zero prefixed numbers as either Strings or Numbers
* Correctly handle nested elements with same name as an ascendant element

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,6 +1,6 @@
 name := "spark-xml"
 
-version := "0.10.0"
+version := "0.10.0.3-SNAPSHOT"
 
 organization := "com.databricks"
 
--- a/scalastyle-config.xml
+++ b/scalastyle-config.xml
@@ -54,7 +54,7 @@ This file is divided into 3 sections:
 
   <check level="error" class="org.scalastyle.file.FileLineLengthChecker" enabled="true">
     <parameters>
-      <parameter name="maxLineLength"><![CDATA[100]]></parameter>
+      <parameter name="maxLineLength"><![CDATA[160]]></parameter>
       <parameter name="tabSize"><![CDATA[2]]></parameter>
       <parameter name="ignoreImports">true</parameter>
     </parameters>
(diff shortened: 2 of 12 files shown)


build.sbt
```diff
@@ -1,6 +1,6 @@
 name := "spark-xml"
 
-version := "0.10.0"
+version := "0.10.0.3-SNAPSHOT"
```
srowen, line 3: This would have to be reverted

scalastyle-config.xml
```diff
@@ -54,7 +54,7 @@ This file is divided into 3 sections:
 
   <check level="error" class="org.scalastyle.file.FileLineLengthChecker" enabled="true">
     <parameters>
-      <parameter name="maxLineLength"><![CDATA[100]]></parameter>
+      <parameter name="maxLineLength"><![CDATA[160]]></parameter>
```
srowen, line 57: And this

src/main/scala/com/databricks/spark/xml/XmlPath.scala
```diff
@@ -0,0 +1,38 @@
+/**
```
srowen, line 1: This would have to be the standard ALv2 header

src/main/scala/com/databricks/spark/xml/XmlRelation.scala
```diff
@@ -18,12 +18,11 @@ package com.databricks.spark.xml
 import java.io.IOException
 
 import org.apache.hadoop.fs.Path
-
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql._
-import org.apache.spark.sql.sources.{PrunedScan, InsertableRelation, BaseRelation, TableScan}
+import org.apache.spark.sql.sources.{ BaseRelation, InsertableRelation, PrunedScan }
```
srowen, line 23: Remove whitespace changes

PR #465, opened 2020-08-12
support include statement with relative path

Gfeuillen:
The current parser does not support specifying the base URI, as a consequence relative imports fail. 
In this PR:

- Set BaseURI when reading xsd schemas

### DIFF
--- a/src/main/scala/com/databricks/spark/xml/util/XSDToSchema.scala
+++ b/src/main/scala/com/databricks/spark/xml/util/XSDToSchema.scala
@@ -39,8 +39,11 @@ object XSDToSchema {
    */
   @Experimental
   def read(xsdFile: File): StructType = {
-    val xmlSchema = new XmlSchemaCollection().read(
+    val xmlSchemaCollection = new XmlSchemaCollection()
+    xmlSchemaCollection.setBaseUri(xsdFile.getParent)
+    val xmlSchema = xmlSchemaCollection.read(
       new InputStreamReader(new FileInputStream(xsdFile), StandardCharsets.UTF_8))
+
     getStructType(xmlSchema)
   }
 
--- a/src/test/resources/include-example/first.xsd
+++ b/src/test/resources/include-example/first.xsd
@@ -1,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema">
+    <xs:include schemaLocation="second.xsd" />
+    <xs:element name="basket" type="basket"/>
+</xs:schema>
--- a/src/test/resources/include-example/second.xsd
+++ b/src/test/resources/include-example/second.xsd
@@ -1,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema">
+    <xs:complexType name="basket">
+        <xs:sequence>
+            <xs:element name="entry" minOccurs="0" maxOccurs="unbounded">
+                <xs:complexType>
+                    <xs:sequence>
+                        <xs:element name="key" minOccurs="0" type="xs:anyType"/>
+                        <xs:element name="value" minOccurs="0" type="xs:anyType"/>
+                    </xs:sequence>
+                </xs:complexType>
+            </xs:element>
+        </xs:sequence>
+    </xs:complexType>
+</xs:schema>
--- a/src/test/scala/com/databricks/spark/xml/util/XSDToSchemaSuite.scala
+++ b/src/test/scala/com/databricks/spark/xml/util/XSDToSchemaSuite.scala
@@ -38,4 +38,19 @@ class XSDToSchemaSuite extends AnyFunSuite {
     assert(expectedSchema === parsedSchema)
   }
 
+  test("Relative path parsing") {
+    val parsedSchema = XSDToSchema.read(
+      Paths.get("src/test/resources/include-example/first.xsd"))
+    val expectedSchema = StructType(Array(
+      StructField("basket", StructType(Array(
+        StructField("entry", ArrayType(
+          StructType(Array(
+            StructField("key", StringType),
+            StructField("value", StringType)
+          )))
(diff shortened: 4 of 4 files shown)


src/main/scala/com/databricks/spark/xml/util/XSDToSchema.scala
```diff
@@ -35,11 +35,14 @@ object XSDToSchema {
...
    * @param xsdFile XSD file
+   * @param baseURI base URI for relative imports
    * @return Spark-compatible schema
    */
   @Experimental
-  def read(xsdFile: File): StructType = {
-    val xmlSchema = new XmlSchemaCollection().read(
+  def fromFile(xsdFile: File, baseURI: Option[String] = None): StructType = {
+    val xmlSchemaCollection = new XmlSchemaCollection()
+    xmlSchemaCollection.setBaseUri(baseURI.getOrElse(xsdFile.getParent))
```
srowen, line 44: When would it not just be xsdFile.getParent?

src/main/scala/com/databricks/spark/xml/util/XSDToSchema.scala
```diff
@@ -35,11 +35,14 @@ object XSDToSchema {
    * Reads a schema from an XSD file.
    *
    * @param xsdFile XSD file
+   * @param baseURI base URI for relative imports
    * @return Spark-compatible schema
    */
   @Experimental
-  def read(xsdFile: File): StructType = {
-    val xmlSchema = new XmlSchemaCollection().read(
+  def fromFile(xsdFile: File, baseURI: Option[String] = None): StructType = {
```
srowen, line 42: Though it's "Experimental" I'd prefer not changing the method names just for its own sake

Gfeuillen: Scala compiler does not allow for having default arguments on multiple overloaded methods

src/main/scala/com/databricks/spark/xml/util/XSDToSchema.scala
Gfeuillen: If the file you read is not at the root of your tree. E.g. if you decide to read a sub part of your tree that references something external

src/main/scala/com/databricks/spark/xml/util/XSDToSchema.scala
srowen: Ah hm OK. We can just add a new method too, which even helps preserve binary compatibility. I'm neutral on it I suppose.

Gfeuillen: What does binary compatibility has to do with this ? 
On the other hand this is still marked `Experimental` and not even released. If you are looking to change it, now might be the best time :)

srowen: The one-arg File method would go away entirely so apps that link against that would fail. Source compatibility is fine though. 
But yes oh I forgot this isn't even released! So I agree, renaming is entirely fine.

src/main/scala/com/databricks/spark/xml/util/XSDToSchema.scala
srowen: I'd assume external resources have to be absolute anyway, but for the relative case, so you want to read `path/bar.xsd` from `foo.xsd` but `path/bar.xsd` doesn't exist at the relative path? that seems wrong. Is the case something like, you copied supporting XSDs to some location for deployment purposes that don't match how they're normally structured?

Gfeuillen: I'm not saying this is the standard use case, hence the sane default.

srowen: I'm not against supporting niche use cases; it's easy here. I'm just trying to understand how it arises at all

Gfeuillen: I don't see any good reason to disallow this, do you want me to change something ? If not feel free to merge :)

srowen: Eh, I think that's a little backwards - we'd add features because they solve a problem not because nobody can think of why not. I'm sure there is a use case for it, but, what is it? can you give any more real world example?

Gfeuillen: This is not my use case, I agree to not make it complicated for ourselves. I will revert that, do you still want to keep the `.read` overloaded method then ? I'd rather not given the limitation mentioned in the other comment.

srowen: OK, if we don't know a use for it, let's not do it yet. I'm open to the idea, just want to make sure it's not being added when there's a better/easier way.
I'm OK with renaming but is there then a reason left to do it? it's also solvable with another overload and no matter what there just won't be many overloads of this or args to methods anyway

Gfeuillen: Only did minimal changes then

src/main/scala/com/databricks/spark/xml/util/XSDToSchema.scala
```diff
@@ -39,8 +39,11 @@ object XSDToSchema {
    */
   @Experimental
   def read(xsdFile: File): StructType = {
-    val xmlSchema = new XmlSchemaCollection().read(
+    val xmlSchemaCollection = new XmlSchemaCollection()
+    xmlSchemaCollection.setBaseUri(xsdFile.getParent)
```
srowen, line 43: Is this not the default? I'm fine with it if it helps.

Gfeuillen: No, this is the whole point of this PR :)

srowen: OK weird I would have thought it assumes that somehow. The change is OK.

Issue #450, opened 2020-06-02
Data loss when input file partitioned through rowTag element


PeterNmp:
Hi,

Thanks for all the effort put into this library!
We still seem to be having this issue related to #399 with 0.9.0 :(
We have large xmlfiles - 10+ GB with format like this:

```
...
<SoundRecording>
...
</SoundRecording>
...
<Release>
...
</Release>
...
<ReleaseTransactions>
...
</ReleaseTransactions>
```

When I count the number of SoundRecording/Release/ReleaseTransactions in the files it is the same (and should be), but processing the files like this:
spark.read.format("com.databricks.spark.xml").....option("rowTag","SoundRecording")
Gives me different counts of SoundRecording/Release/ReleaseTransactions for some files processed.

srowen:
Are you sure there are the same amounts? hard to say without any reproduction
Compressed or uncompressed?

PeterNmp:
They are uncompressed. I can try process the compressed .gz files (that did help when we had the problem before version 0.7.0).
I'm fairly sure about the counts - did count like this on the files:
grep -o '<Release>' file.xml | wc -l
Count here didn't match result form processing in databricks.
btw we are running with "option("mode","FAILFAST")"

PeterNmp:
Sorry counts like this:
grep -o '<Release>' file.xml | wc -l

srowen:
Yeah I'd be interested if the compressed case is different. They are different code paths and both rely a bit on assumptions about the implementation to get it right. The main fix last time was for the uncompressed path indeed.
What Hadoop / Spark version?

PeterNmp:
We are running Databricks in Azure 6.3 (includes Apache Spark 2.4.4, Scala 2.11)
Tried running on an uncompressed file - it gives the correct count on all the dataframes.
When running on the same uncompressed file one of the dataframes consistently gets the wrong count - one less than expected.
I'm sorry but i can't share the files :(

srowen:
Hm, OK. What kind of compression? I do have some tests that check compressed files across block boundaries but there may well be all kinds of corner cases. Really, splittable or unsplittable compression?

PeterNmp:
It's gzip compression. I don't know if it's splittable or not but the compressed files seem to run slower and require more memory on the nodes.
Before the fix in 0.7.0 we saw the same - that the gzipped files would process correctly but not the unzipped ones.

srowen:
[...]

PR #467, opened 2020-08-16
[WIP] Assume unsplittable codec means partition is whole file; don't use compressed input file position vs end

srowen:
See https://github.com/databricks/spark-xml/issues/450

This may be the source of the data loss issue when reading unsplittably-compressed XML.
I think the problem is deciding when the partition is exhausted by looking at the underlying compressed stream's input position. It will have been read ahead by buffering, so, may be beyond partition end before it's consumed.

For unsplittable compression, the whole file is the partition. I believe that means we can hack a little more and just ignore the position vs end in this case. As long as more data is available, keep reading.

I'm 95% sure about that but still makes me nervous.
Existing compressed input tests pass at least.
I'd love to get this tested by something that can reproduce it.


srowen:
CC @PeterNmp - would you be able to try a build with this change.
CC @HyukjinKwon FYI, no action needed.

PR #468, opened 2020-08-18
Take into account StreamDecoder.hasLeftoverChar in trying to exactly always correctly determine how much has been read

srowen:
See #450 

The idea here is that we have to account for exactly how much data has been read from the underlying split, but not yet evaluated. Buffering is the problem. Previously we accounted for data buffered in StreamDecoder's ByteBuffer. I think we may have to account for the fact that it reads 2 chars at a time, and so may have buffered one more.

It should be extremely rare that this causes a problem - it would have to have read a start `<` that is the final char in a split, and have no buffered last char. But it's possible and indeed see the issue above - in tens of millions of records it might happen.

I don't know if this is the solution, but an OK guess.

### DIFF
--- a/src/main/scala/com/databricks/spark/xml/XmlInputFormat.scala
+++ b/src/main/scala/com/databricks/spark/xml/XmlInputFormat.scala
@@ -65,6 +65,7 @@ private[xml] class XmlRecordReader extends RecordReader[LongWritable, Text] {
   private var reader: Reader = _
   private var filePosition: Seekable = _
   private var countingIn: CountingInputStream = _
+  private var readerLeftoverCharFn: () => Boolean = _
   private var readerByteBuffer: ByteBuffer = _
   private var decompressor: Decompressor = _
   private var buffer = new StringBuilder()
@@ -127,6 +128,9 @@ private[xml] class XmlRecordReader extends RecordReader[LongWritable, Text] {
       val sdField = reader.getClass.getDeclaredField("sd")
       sdField.setAccessible(true)
       val sd = sdField.get(reader)
+      val readerLeftoverCharField = sd.getClass.getDeclaredField("haveLeftoverChar")
+      readerLeftoverCharField.setAccessible(true)
+      readerLeftoverCharFn = () => { readerLeftoverCharField.get(sd).asInstanceOf[Boolean] }
       val bbField = sd.getClass.getDeclaredField("bb")
       bbField.setAccessible(true)
       readerByteBuffer = bbField.get(sd).asInstanceOf[ByteBuffer]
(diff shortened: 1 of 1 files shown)


src/main/scala/com/databricks/spark/xml/XmlInputFormat.scala
```diff
@@ -149,7 +153,9 @@ private[xml] class XmlRecordReader extends RecordReader[LongWritable, Text] {
     if (filePosition != null) {
       filePosition.getPos
     } else {
-      start + countingIn.getByteCount - readerByteBuffer.remaining()
+      start + countingIn.getByteCount -
+        readerByteBuffer.remaining() -
+        (if (readerLeftoverCharFn()) 1 else 0)
```
srowen, line 158: 1 buffered char doesn't necessarily mean one byte. I don't see a way to reason about how many bytes the char actually caused to be read. Because we are worried about the case of `<`, which is a single byte in most encodings, this is mostly OK. I truly hope that Hadoop figures out how to not split multibyte chars across splits, or else, I can't see how any multibyte text is reliably read when split by Hadoop.

srowen:
@ericsun95 @PeterNmp not sure if either of you have time to try a build from this branch to see if it fixes your issue. I can create an assembly JAR if needed. You're my only known reproducers!

ericsun95:
> @ericsun95 @PeterNmp not sure if either of you have time to try a build from this branch to see if it fixes your issue. I can create an assembly JAR if needed. You're my only known reproducers!

I can have a try. Could you show me how to build the jar on top of this / give this assembly JAR to me?

srowen:
@ericsun95 sure, are you on Scala 2.11 or 2.12?

ericsun95:
> @ericsun95 sure, are you on Scala 2.11 or 2.12?

2.11

srowen:
@ericsun95 Try this assembly JAR. Be sure to remove any other versions you have first:
https://drive.google.com/file/d/13rPJoH814VycIbBbLWfPCzFvoJ-KXzT-/view?usp=sharing

PeterNmp:
@srowen - thanks i'll test it too! Do i need to do anything apart from installing the assembly?

PeterNmp:
It looks really good!!! I've tested two files that fails with 0.9.0.
Loaded both compressed and uncompressed file. Same number of rows in dataframes. Subtracting uncompressed from compressed and vice versa gives no result. Looking forward to this being released! Again - thanks for the effort!

srowen:
Good to hear that was a good guess. If there are no other comments I'll commit and roll a new release shortly. Thank you for testing.

ericsun95:
> Good to hear that was a good guess. If there are no other comments I'll commit and roll a new release shortly. Thank you for testing.

The number is correct. While when I repartition, it will throw null pointer exception, not sure if there is data corruption during this process, will have more test to check.

srowen:
If the NPE is from spark-xml, post it, and I'll try to figure out if it's related or another issue.

ericsun95:
> If the NPE is from spark-xml, post it, and I'll try to figure out if it's related or another issue.

Congs. It's just a small bug in my code. I think this solution works well! Thanks!

srowen:
FYI I just released 0.10.0 with this and other changes / fixes. 
https://github.com/databricks/spark-xml/releases/tag/v0.10.0

PR #473, opened 2020-09-01
Adding header

rtjarvis:
Updated following review.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,8 @@ This package allows reading XML files in local or distributed filesystem as [Spa
 When reading files the API accepts several options:
 
 * `path`: Location of files. Similar to Spark can accept standard Hadoop globbing expressions.
-* `rowTag`: The row tag of your xml files to treat as a row. For example, in this xml `<books> <book><book> ...</books>`, the appropriate value would be `book`. Default is `ROW`.
+* `rowTag`: The row tag of your xml files to treat as a row. For example, in this xml `<books> <book><book> ...</books>`, the appropriate value would be `book`. Default is `ROW`. An empty rootTag will omit a root tag entirely
+* `header`: A raw header to add at the top of the output file. Can be used to add document encoding declaration
 * `samplingRatio`: Sampling ratio for inferring schema (0.0 ~ 1). Default is 1. Possible types are `StructType`, `ArrayType`, `StringType`, `LongType`, `DoubleType`, `BooleanType`, `TimestampType` and `NullType`, unless user provides a schema for this.
 * `excludeAttribute` : Whether you want to exclude attributes in elements or not. Default is false.
 * `treatEmptyValuesAsNulls` : (DEPRECATED: use `nullValue` set to `""`) Whether you want to treat whitespaces as a null value. Default is false
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ This package allows reading XML files in local or distributed filesystem as [Spa
 When reading files the API accepts several options:
 
 * `path`: Location of files. Similar to Spark can accept standard Hadoop globbing expressions.
-* `rowTag`: The row tag of your xml files to treat as a row. For example, in this xml `<books> <book><book> ...</books>`, the appropriate value would be `book`. Default is `ROW`. An empty rootTag will omit a root tag entirely
+* `rowTag`: The row tag of your xml files to treat as a row. For example, in this xml `<books> <book><book> ...</books>`, the appropriate value would be `book`. Default is `ROW`.
 * `header`: A raw header to add at the top of the output file. Can be used to add document encoding declaration
 * `samplingRatio`: Sampling ratio for inferring schema (0.0 ~ 1). Default is 1. Possible types are `StructType`, `ArrayType`, `StringType`, `LongType`, `DoubleType`, `BooleanType`, `TimestampType` and `NullType`, unless user provides a schema for this.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -84,7 +84,8 @@ When writing files the API accepts several options:
 
 * `path`: Location to write files.
 * `rowTag`: The row tag of your xml files to treat as a row. For example, in this xml `<books> <book><book> ...</books>`, the appropriate value would be `book`. Default is `ROW`.
-* `rootTag`: The root tag of your xml files to treat as the root. For example, in this xml `<books> <book><book> ...</books>`, the appropriate value would be `books`. Default is `ROWS`.
+* `rootTag`: The root tag of your xml files to treat as the root. For example, in this xml `<books> <book><book> ...</books>`, the appropriate value would be `books`. Default is `ROWS`. An empty rootTag will omit a root tag entirely.
+* `header`: A raw header to add at the top of the output file. Can be used to add document encoding declaration. Can be used to hold an XML declaration eg: `<?xml version="1.0" encoding="UTF-8"?>`
```
srowen, line 88: Could add something like "Available from version 0.11.0 onwards" here?

src/main/scala/com/databricks/spark/xml/util/XmlFile.scala
```diff
@@ -93,7 +93,12 @@ private[xml] object XmlFile {
         override def next: String = {
           if (iter.nonEmpty) {
             if (firstRow) {
-              indentingXmlWriter.writeStartElement(options.rootTag)
+              if (options.header != null) {
+                writer.append(options.header)
+              }
+              if (!options.rootTag.isEmpty) {
```
srowen, line 99: I don't think we can escape writing the root tag. It would not be valid XML otherwise.

src/main/scala/com/databricks/spark/xml/util/XmlFile.scala
```diff
@@ -93,7 +93,12 @@ private[xml] object XmlFile {
         override def next: String = {
           if (iter.nonEmpty) {
             if (firstRow) {
-              indentingXmlWriter.writeStartElement(options.rootTag)
+              if (options.header != null) {
+                writer.append(options.header)
```
srowen, line 97: You might have to write a newline here too

src/test/scala/com/databricks/spark/xml/XmlSuite.scala
```diff
@@ -474,6 +475,26 @@ final class XmlSuite extends AnyFunSuite with BeforeAndAfterAll {
     assert(booksCopy.collect.map(_.toString).toSet === books.collect.map(_.toString).toSet)
   }
 
+  test("DSL save with header defined") {
```
srowen, line 478: I apologize, I think you'll have to rebase this PR. I merged a change just now that touches the same code.
But as an upshot, there's a test case I just added which you could copy to test the first line of the output - I think that's really what you want to test, not just that it reads back. Because that would work even if this didn't write a header.
See `  test("rootTag with simple attributes") {`

src/main/scala/com/databricks/spark/xml/util/XmlFile.scala
huejim: Depends on schema. Root tag is required when exporting a simple table. In case of a complex schema e.g.:
```
StructType(
  StructField("person",
    ArrayType(
      StructType(...)
    )),
  StructField("animal",
    ArrayType(
      StructType(...)
    ))
)
```
where the table has a single row with array columns the produced xml has an extra root tag:
```
<root>
  <row>
    <person> ... </person>
    <person> ... </person>
    <animal> ... </animal>
    <animal> ... </animal>
  </row>
<root>
```
This is definitely a useful feature.

srowen: I don't think so. You would have a doc like:

```
  <row>
    <person> ... </person>
    <person> ... </person>
    <animal> ... </animal>
    <animal> ... </animal>
  </row>
  <row>
    <person> ... </person>
    <person> ... </person>
    <animal> ... </animal>
    <animal> ... </animal>
  </row>
  ...
```
There would be no root element in the XML file.

huejim: My mistake, it doesn't depend on schema complexity but on number of rows.

Table with only one row produces a single row element therefore it can act as the root element. I know that my table has a single row because it's constructed that way: multiple tables are joined together and collapsed into arrays of structs preserving their hierarchy.

Maybe there should be some check throwing an exception if the source table has multiple rows.

srowen: OK, I don't really object. It won't be the default and hey if someone really wants to do it, I guess they can if they know what they're doing.

src/main/scala/com/databricks/spark/xml/util/XmlFile.scala
```diff
@@ -109,7 +114,9 @@ private[xml] object XmlFile {
           } else {
             if (!firstRow) {
               lastRow = false
-              indentingXmlWriter.writeEndElement()
+              if (!options.rootTag.isEmpty) {
```
srowen, line 117: total nit but you can write just `.nonEmpty`

srowen (commented):
If you're willing to make a few more minor changes in the comments, I think it's OK.

srowen:
Feel free to reopen with a rebase etc

PR #475, opened 2020-09-03
Fix error in parsing XSD types; handle attributes on all elements; more correctly handle nullability in XSDs

### LINKED ISSUE #474: XSD types not parsed correctly in some cased in XSDToSchema + XSD attributes ignored in some cases

srowen:
Closes #474 

- Fixes how QNames were looked up from the schema - did not actually work before
- Attributes were not handled in most cases; now they are
- Nullability wasn't handled correctly in many cases according to minOccurs; now should be
- Refactored test code to share schema construction code

### DIFF
--- a/src/main/scala/com/databricks/spark/xml/util/XSDToSchema.scala
+++ b/src/main/scala/com/databricks/spark/xml/util/XSDToSchema.scala
@@ -121,47 +121,47 @@ object XSDToSchema {
             content.getContent match {
               case extension: XmlSchemaSimpleContentExtension =>
                 val baseStructField = getStructField(xmlSchema,
-                  xmlSchema.getTypeByName(extension.getBaseTypeName))
+                  xmlSchema.getParent.getTypeByQName(extension.getBaseTypeName))
                 val value = StructField("_VALUE", baseStructField.dataType)
                 val attributes = extension.getAttributes.asScala.map {
                   case attribute: XmlSchemaAttribute =>
                     val baseStructField = getStructField(xmlSchema,
-                      xmlSchema.getTypeByName(attribute.getSchemaTypeName))
-                    StructField(s"_${attribute.getName}", baseStructField.dataType)
+                      xmlSchema.getParent.getTypeByQName(attribute.getSchemaTypeName))
+                    StructField(s"_${attribute.getName}", baseStructField.dataType,
+                      attribute.getUse != XmlSchemaUse.REQUIRED)
                 }
                 StructField(complexType.getName, StructType(value +: attributes))
             }
-          case null =>
-            complexType.getParticle match {
-              // xs:all
-              case all: XmlSchemaAll =>
-                val fields = all.getItems.asScala.map {
-                  case element: XmlSchemaElement =>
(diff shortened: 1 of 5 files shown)


src/main/scala/com/databricks/spark/xml/util/XSDToSchema.scala
```diff
@@ -121,64 +121,73 @@ object XSDToSchema {
             content.getContent match {
               case extension: XmlSchemaSimpleContentExtension =>
                 val baseStructField = getStructField(xmlSchema,
-                  xmlSchema.getTypeByName(extension.getBaseTypeName))
+                  xmlSchema.getParent.getTypeByQName(extension.getBaseTypeName))
```
srowen, line 124: This fixes the lookup; xmlSchema has no built in types to lookup but the parent collection does.

src/main/scala/com/databricks/spark/xml/util/XSDToSchema.scala
```diff
@@ -121,64 +121,73 @@ object XSDToSchema {
...
-                      field
+                      StructField(element.getName, baseStructField.dataType, nullable)
                     } else {
-                      StructField(element.getName, ArrayType(field.dataType))
+                      StructField(element.getName, ArrayType(baseStructField.dataType), nullable)
                     }
                   }
-                StructField(complexType.getName, StructType(fields))
+              }
+            val attributes = complexType.getAttributes.asScala.map {
```
srowen, line 179: This adds attribute handling on other element types

src/main/scala/com/databricks/spark/xml/util/XSDToSchema.scala
```diff
@@ -121,64 +121,73 @@ object XSDToSchema {
...
+                      } else {
+                        StructField(element.getName, ArrayType(baseStructField.dataType), nullable)
+                      }
+                  }
+                // xs:choice
+                case choice: XmlSchemaChoice =>
+                  choice.getItems.asScala.map { case element: XmlSchemaElement =>
                     val baseStructField = getStructField(xmlSchema, element.getSchemaType)
-                    val field = StructField(element.getName, baseStructField.dataType)
+                    val nullable = element.getMinOccurs == 0
```
srowen, line 154: Several fixes like this to account for min occurs in deciding nullability.

src/test/scala/com/databricks/spark/xml/TestUtils.scala
```diff
@@ -0,0 +1,35 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.databricks.spark.xml
+
+import org.apache.spark.sql.types.{ArrayType, DataType, StringType, StructField, StructType}
+
+private[xml] object TestUtils {
```
srowen, line 19: Pulled this code out to share the schema construction utilities.

srowen (commented):
CC @seddonm1 in case you're interested.
CC @miku872 too

PR #476, opened 2020-09-08
Remove spark-packages config; minor code updates; update scalatest et al

srowen:


srowen:
Hm, not sure why Travis CI is not providing build results here, but it passed: https://travis-ci.org/github/databricks/spark-xml/builds/725281434
I was messing with Travis CI in other projects yesterday but don't think I can affect spark-xml, weird.

PR #479, opened 2020-09-16
extended support for XmlSchemaComplexContentExtension in XSDToSchema.getStructField

miku872:

PR #482, opened 2020-09-17
extended support for XmlSchemaComplexContentExtension in XSDToSchema.getStructField

miku872:
Extended support for XmlSchemaComplexContentExtension and changed the way extension elements were being added to baseStructField

### DIFF
--- a/src/main/scala/com/databricks/spark/xml/util/XSDToSchema.scala
+++ b/src/main/scala/com/databricks/spark/xml/util/XSDToSchema.scala
@@ -70,7 +70,6 @@ object XSDToSchema {
     getStructType(xmlSchema)
   }
 
-
   private def getStructField(xmlSchema: XmlSchema, schemaType: XmlSchemaType): StructField = {
     schemaType match {
       // xs:simpleType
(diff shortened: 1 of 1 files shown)


srowen:
How does this differ from #479 ?

miku872:
Hey Srowen, 
I changed the way baseStructFields were being added to extension elements. Previously baseStructFields were added as a separate node. In this PR baseStructField elements are being added directly instead of being added as separate node. I hope that makes sense. Also I refactored some code.

src/main/scala/com/databricks/spark/xml/util/XSDToSchema.scala
```diff
@@ -23,13 +23,15 @@ import org.apache.spark.annotation.Experimental
...
 import org.apache.ws.commons.schema.constants.Constants
+import scala.collection.mutable
 
 /**
  * Utility to generate a Spark schema from an XSD. Not all XSD schemas are simple tabular schemas,
  * so not all elements or XSDs are supported.
  */
 @Experimental
 object XSDToSchema {
+  val schemaNameToStructFieldMap = collection.mutable.Map[String, StructField]()
```
srowen, line 34: You can just write mutable.Map
This can't be a global mutable value though

srowen:
Closed in favor of #483

PR #483, opened 2020-09-25
extended support for  XSDToSchema

miku872:
I have added support for complexContent extension, element references, choice, group, group references, attribute group and references etc.


srowen:
Before we get too far into it, do you have tests to check the new functionality?

miku872:
yes, i checked with at least 10 different xsd which contain the cases i
covered

miku872:
I'll add the test cases

PR #486, opened 2020-09-30
FIX #485 Fix Timestamp to conform to XML Spec

### LINKED ISSUE #485: Standardise Writer Formats
Hi,
When we did the work on https://github.com/databricks/spark-xml/pull/448 we should have also done the writing side to ensure the writer would conform with the spec.

We are currently (https://github.com/databricks/spark-xml/blob/master/src/main/scala/com/databricks/spark/xml/parsers/StaxXmlGenerator.scala#L79) relying on the default `java.sql.Timestamp` format (https://docs.oracle.com/javase/8/docs/api/java/sql/Timestamp.html#toString--) which does not define timezone properly.

I think it would be good to switch to a more deterministic format:

```scala
writer.writeCharacters(DateTimeFormatter.ISO_INSTANT.format(v.toInstant()))
```

seddonm1:
We are currently (https://github.com/databricks/spark-xml/blob/master/src/main/scala/com/databricks/spark/xml/parsers/StaxXmlGenerator.scala#L79) relying on the default java.sql.Timestamp format (https://docs.oracle.com/javase/8/docs/api/java/sql/Timestamp.html#toString--) which formats timestamp objects in local time whereas the spec https://www.w3.org/TR/xmlschema-2/#dateTime indicates everything should be in UTC.

As the spec supports the inclusion of the timezone 'Z' we should be explicit so anyone manually reading a file can also understand the accurate timestamp.

See https://github.com/databricks/spark-xml/issues/485

### DIFF
--- a/src/main/scala/com/databricks/spark/xml/parsers/StaxXmlGenerator.scala
+++ b/src/main/scala/com/databricks/spark/xml/parsers/StaxXmlGenerator.scala
@@ -15,6 +15,8 @@
  */
 package com.databricks.spark.xml.parsers
 
+import java.time.format.DateTimeFormatter
+
 import javax.xml.stream.XMLStreamWriter
 
 import scala.collection.Map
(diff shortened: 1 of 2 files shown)


srowen (approved):
Looks good to me if tests pass.

seddonm1:
I have added a 'roundtrip' test that confirms that what we write with this plugin can be read back without losing information (which was not true previously unless your localtime = UTC).

srowen:
Weird, Travis CI tests aren't triggering anymore. I can test manually while I figure it out.

srowen:
Looks OK locally so I'll merge

Issue #484, opened 2020-09-28
Schema difference


trourance:
At the moment, it's not possible to generate a schema with a xsd file and read a xml file using the previously generated schema.
Result to do so will give you a runtime exception `The 0th field '...' of input row cannot be null.`
Trying to infer schema directly from xml file will also give you a different result than the one generated from xsd.
I tried the 'shiporder' example at https://www.w3schools.com/xml/schema_example.asp
Schema generated from xsd:
```
root
 |-- shiporder: struct (nullable = false)
 |    |-- orderperson: string (nullable = false)
 |    |-- shipto: struct (nullable = false)
 |    |    |-- name: string (nullable = false)
 |    |    |-- address: string (nullable = false)
 |    |    |-- city: string (nullable = false)
 |    |    |-- country: string (nullable = false)
 |    |-- item: array (nullable = false)
 |    |    |-- element: struct (containsNull = true)
 |    |    |    |-- title: string (nullable = false)
 |    |    |    |-- note: string (nullable = true)
 |    |    |    |-- quantity: integer (nullable = false)
 |    |    |    |-- price: decimal(38,18) (nullable = false)
 |    |-- _orderid: string (nullable = false)
```

 Schema inferred from xml using `rowTag` 'shiporder':
```
root
 |-- _orderid: long (nullable = true)
 |-- _xmlns:xsi: string (nullable = true)
 |-- _xsi:noNamespaceSchemaLocation: string (nullable = true)
 |-- item: array (nullable = true)
 |    |-- element: struct (containsNull = true)
 |    |    |-- note: string (nullable = true)
 |    |    |-- price: double (nullable = true)
 |    |    |-- quantity: long (nullable = true)
 |    |    |-- title: string (nullable = true)
 |-- orderperson: string (nullable = true)
 |-- shipto: struct (nullable = true)
 |    |-- address: string (nullable = true)
 |    |-- city: string (nullable = true)
 |    |-- country: string (nullable = true)
 |    |-- name: string (nullable = true)
```

srowen:
Yes, this doesn't 100% work in many cases.

In this example though, you will need to pull out the schema of `shiporder` to use to parse, at least. That is really the structure of the row tag, not a child of it. That looks roughly 'correct' then.

What I'm not sure of is what it will do with the xmlns and xsi attributes. If they're not in the schema then it may not like parsing those, even though they are kind of metadata. If that's an issue we may have to look at special-casing some namespaces to ignore optionally.

[...]

PR #487, opened 2020-09-30
Allow multiple root elements in XSD

srowen:
This relates to issue #484 

The XSD -> schema tool currently only pays attention to the first element in the XSD. There's no good reason not to allow multiple ones.

I also believe I fixed a possible bug in determining the element name.

### DIFF
--- a/src/main/scala/com/databricks/spark/xml/util/XSDToSchema.scala
+++ b/src/main/scala/com/databricks/spark/xml/util/XSDToSchema.scala
@@ -33,6 +33,9 @@ object XSDToSchema {
 
   /**
    * Reads a schema from an XSD file.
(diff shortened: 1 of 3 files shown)


trourance:
As far as I know, defining multiple root elements is not authorized in XML, nor in XSD.

See https://www.w3schools.com/xml/schema_intro.asp and https://www.w3schools.com/xml/xml_syntax.asp, specifically:

- it must have one unique root element

srowen:
Yes XML can't have multiple root elements, but here we're defining the schema of a child element.
The XSD still defines one schema and has one root of course (it's XML itself), but it is defining the schema of an element that can have multiple children. (So 'root' isn't the best choice of words above maybe, yes)
At least, the XSD parser is fine with it!

PR #491, opened 2020-10-05
Assume xs:choice elements are nullable

### LINKED ISSUE #490: xs:choice element can't be null
When the schema is generated from a XSD using XSDToSchema, fields resulting from xs:element included in xs:choice have the `nullable` property set to `false`, unless attribute `MinOccurs == 0`.
IMO, this is a wrong assumption because the XML file will always have one element OR the other, and therefore one will be null.
Wouldn't it be more correct to set `nullable` as always `true` in those resulting fields ?

srowen:
Closes #490 
Because only one child of an xs:choice can actually appear, those fields should be inherently nullable always.


srowen:
CC @trourance

trourance:
It works perfectly, thanks.

PR #492, opened 2020-10-08
Update to sbt 1.2.x (1.3+ not working yet); fix latent scalastyle issue

srowen:
Not sure why, but different classpath treatment during tests in sbt 1.3 makes Spark 2.x tests fail. 
Otherwise just fixing a breakage in master while updating to at least SBT 1.2.8

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 target/
 .idea/
 .idea_modules/
+.bsp/
 .DS_Store
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ jobs:
       env:
         - TEST_SPARK_VERSION="2.3.4"
     - env:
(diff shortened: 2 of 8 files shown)


HyukjinKwon (approved):
Nice!

srowen:
I don't think Travis CI PR builds are populating here anymore for some reason; here it is: https://travis-ci.org/github/databricks/spark-xml/builds/733981411

I think that the StaxXmlGeneratorSuite was not actually being run (!) possibly because its package was wrong. I cleaned it up a bit and it works.

I'm also not totally clear why the scalastyle errors weren't being picked up by the PR builds before merging to master. Probably missed this in the confusion of not finding the PR results linked here.

The sbt 1.3+ error concerns Guava; somehow there's a version conflict in the classpath when building for Spark 2.x but not 3.x.

Issue #480, opened 2020-09-16
xs:any support


trourance:
Hi,
I was trying to create a schema from a XSD using XSDToSchema. It failed with the following error:
```
scala.MatchError: org.apache.ws.commons.schema.XmlSchemaAny@afdf919f (of class org.apache.ws.commons.schema.XmlSchemaAny)
  at com.databricks.spark.xml.util.XSDToSchema$.$anonfun$getStructField$4(XSDToSchema.scala:165)
```
I've found it's because the XSD has this complexType defined in it:
```
<xs:complexType name="SupplementaryDataEnvelope1">
    <xs:sequence>
        <xs:any namespace="##any" processContents="lax"/>
    </xs:sequence>
</xs:complexType>
```
It seems there was a proposal to support this type as string here: https://github.com/databricks/spark-xml/issues/449

Any chance to support process for this type ?

srowen:
That issue was referring to the simple type anyType, which is something else. That's handled as a string, yes.
I don't know XSDs well but this seems to be about namespaces. It doesn't define a type? there is no name, etc - it's not an XSD element. If you want a sequence of anyType you'd write an xs:element in there.

What the parser can do is just ignore things it doesn't recognize like this, at least. I'm not sure if the result would be 'correct' in all cases because handling namespaces is tricky, but at least it wouldn't fail.

That said in your case it would still not be valid I think, as there is nothing inside the sequence?

trourance:
I should have been more precise. This is how is defined the element that uses the complex type where xs:any is included:
```
<xs:complexType name="SupplementaryData1">
    <xs:sequence>
        <xs:element maxOccurs="1" minOccurs="0" name="PlcAndNm" type="Max350Text"/>
        <xs:element name="Envlp" type="SupplementaryDataEnvelope1"/>
    </xs:sequence>
</xs:complexType>
<xs:complexType name="SupplementaryDataEnvelope1">
    <xs:sequence>
        <xs:any namespace="##any" processContents="lax"/>
    </xs:sequence>
</xs:complexType>
```

srowen:
I may just not know enough, but I don't know what the second sequence is meant to be. xs:any doesn't appear to define any type.

trourance:
It seems correct, xs:any doesn't define any type. I understand that you can put whatever you want in it.

https://www.w3schools.com/xml/el_any.asp
https://www.w3schools.com/xml/schema_complex_any.asp

srowen:
[...]

PR #493, opened 2020-10-12
Support xs:any in XSD as a 'wildcard' schema element in XML parser

srowen:
See https://github.com/databricks/spark-xml/issues/480

This attempts to support a special 'wildcard' schema element, called by default "xs_any", that will match any unmatched XML elements in a node, and treat them as strings containing the raw XML content. If it's a string, it will match one element. If it's an array of strings, it will match all of them.

This provides a natural interpretation of "xs:any" in XSDs then.

### DIFF
--- a/src/main/scala/com/databricks/spark/xml/parsers/StaxXmlParser.scala
+++ b/src/main/scala/com/databricks/spark/xml/parsers/StaxXmlParser.scala
@@ -294,6 +294,10 @@ private[xml] object StaxXmlParser extends Serializable {
     convertAttributes(rootAttributes, schema, options).toSeq.foreach { case (f, v) =>
       nameToIndex.get(f).foreach { row(_) = v }
     }
+    
+    val wildcardColName = options.wildcardColName
+    val hasWildcard = schema.exists(_.name == wildcardColName)
+    
     var badRecordException: Option[Throwable] = None
 
     var shouldStop = false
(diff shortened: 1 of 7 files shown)


src/main/scala/com/databricks/spark/xml/parsers/StaxXmlParser.scala
```diff
@@ -325,7 +325,23 @@ private[xml] object StaxXmlParser extends Serializable {
             }
 
             case None =>
-              StaxXmlParserUtils.skipChildren(parser)
+              if (schema.exists(_.name == "xs_any")) {
```
srowen, line 328: I suppose we should make "xs_any" configurable

src/test/scala/com/databricks/spark/xml/XmlSuite.scala
```diff
@@ -755,6 +755,39 @@ final class XmlSuite extends AnyFunSuite with BeforeAndAfterAll {
...
+    val schema = buildSchema(
+      field(s"${DEFAULT_ATTRIBUTE_PREFIX}id"),
+      field("author"),
+      field("description"),
+      field("genre"),
+      field("price", DoubleType),
+      field("publish_date"),
+      field("xs_any"))
+    val results = spark.read.schema(schema).option("rowTag", booksTag).xml(booksFile)
+      // .select("xs_any")
```
srowen, line 768: See https://github.com/databricks/spark-xml/issues/480#issuecomment-706338172 for comments about the behavior after select(); not sure if we can change it or if it is OK to leave anyway

src/main/scala/com/databricks/spark/xml/util/XSDToSchema.scala
```diff
@@ -158,29 +158,39 @@ object XSDToSchema {
                   }
                 // xs:choice
                 case choice: XmlSchemaChoice =>
```
srowen, line 160: I don't think xs:any can appear in xs:all so did not add it above

srowen:
@HyukjinKwon not sure if you'd like to review this - not necessary.
My main concern is what I commented on at https://github.com/databricks/spark-xml/issues/480#issuecomment-706338172

srowen:
I'm going to add this. It won't break existing workloads, and if the functionality works oddly in some cases mentioned above, well, it's still good to have it mostly working.

PR #518, opened 2021-02-10
Fix parsing of partial result when corrupted record field is present

### LINKED ISSUE #517: Detection of corrupted rows in XML
Hi all, 

What is the correct way of detecting/handling corrupted records in a XML file? 

We've an XML file that we're trying to parse, it contains data that could be one of the following types: `Double`, `String`, or `Array<Struct>`. We've been successful in parsing the data correctly with our custom schema. However, when we try detecting the corrupted records, by adding a new column to our custom schema called `_corrupt_record` and setting the option `columnNameOfCorruptRecord` to `_corrupt_record`, we get random exceptions like: `The value (1.0) of the type (java.lang.Double) cannot be converted to an array of struct<_VALUE:string, _m:int>`, or `The value (-10000.0) of the type (java.lang.Double) cannot be converted to the string type`.

Also, we've noted that the order of the columns in the resulted Dataframe doesn't match the submitted schema, when we added the `columnNameOfCorruptRecord`, and the new `_corrupt_record` column to our custom schema.

Here's a sample of our XML file data:

```
<row>
    <c1>data1</c1>
    <c2>data2</c2>
    <c2 m=2>data2_m2</c2>
    <c3 m=2>1234</c3>
</row>
```

The corresponding schema to this sample should be the following:
```
root
 |-- c1: string (nullable = true)
 |-- c2: array (nullable = true)
 |    |-- element: struct (containsNull = true)
 |    |    |-- _VALUE: string(nullable = true)
 |    |    |-- _m: integer (nullable = true)
 |-- c3: array (nullable = true)
 |    |-- element: struct (containsNull = true)
 |    |    |-- _VALUE: double(nullable = true)
 |    |    |-- _m: integer (nullable = true)
 ```

### PR DESCRIPTION
**Pull request by srowen**
Closes #517

### CODE REVIEW

File: src/main/scala/com/databricks/spark/xml/parsers/StaxXmlParser.scala [hunk 1]
```diff
@@ -115,16 +115,13 @@ private[xml] object StaxXmlParser extends Serializable {
       case PermissiveMode =>
         logger.debug("Malformed line cause:", cause)
         // The logic below is borrowed from Apache Spark's FailureSafeParser.
-        val corruptFieldIndex = Try(schema.fieldIndex(options.columnNameOfCorruptRecord)).toOption
-        val actualSchema = StructType(schema.filterNot(_.name == options.columnNameOfCorruptRecord))
         val resultRow = new Array[Any](schema.length)
-        var i = 0
-        while (i < actualSchema.length) {
-          val from = actualSchema(i)
-          resultRow(schema.fieldIndex(from.name)) = partialResult.map(_.get(i)).orNull
```
srowen, line 124: @HyukjinKwon this was the bug. I call your attention to it because as the comment above says, this is copied from Spark's CSV parser. If I'm really right about this I need to fix a similar problem in Spark.

The bug is basically: `actualSchema` is the schema without the corrupt record field. We iterate over those (index `i`) and set the field in the partial result according to its real `fieldIndex`. But we read from `partialResult` at `i`, which is an index into `actualSchema`, which has one less field, while `partialResult` has the same schema as the resulting Row. If it's in the middle, then everything after is off by 1 when returned to catalyst.

srowen: I'm tracking a similar fix in Spark in https://issues.apache.org/jira/browse/SPARK-34422

Issue #519, opened 2021-02-14
Setting Timestamp Format as an Option


ma-khalil:
Hi,

Is there a way to set the timestamp format as an option, like we do in reading the `csv` files, the `timestampFormat` option?

I've tried setting it using the following code

```
spark.read
      .option("rowTag", "row")
      .option("timestampFormat", "dd-MM-yyyy")
      .schema(schema)
      .xml("test.xml").show()
```
But it always returns null in the `Timestamp` field. Also, I've tried searching for a method representing that option in the `XMLReader` but I couldn't find any.

It would be really great if you could add setting the `timestamp` as an option while reading the XML files, if it's not already there :)

srowen:
Can you show your schema and sample XML? There is some support for parsing dates, but I am not sure if it's 100% right. Timestamps (with hour/minute/second) should work.

ma-khalil:
> Timestamps (with hour/minute/second) should work.

I've tried adding hour/minute/second with the format proposed above, but it didn't work. Only the default one, with hour/minute/second written, works. Probably, the data that I'd get won't contain hours/minutes/seconds in it.

Here's the XML:

```
<book>
    <author>John Smith</author>
    <date>01-01-2021</date>
</book>
```

Here's the schema as well:

```
root
 |-- author: string (nullable = true)
 |-- date: timestamp (nullable = true)
```

Here's the complete code reading the XML:
```
import com.databricks.spark.xml._
import org.apache.spark.sql.SparkSession
import org.apache.log4j.{Level, Logger}
import org.apache.spark.sql.types.{StringType, StructField, StructType, TimestampType}

object Main {
  def main(args: Array[String]): Unit = {
    Logger.getLogger("org").setLevel(Level.ERROR)
    val spark = SparkSession
      .builder()
      .master("local")
      .getOrCreate()

    val schema = StructType(
      Array(
        StructField("author", StringType),
        StructField("date", TimestampType)
      )
    )

    schema.printTreeString()

    spark.read
      .option("rowTag", "book")
      .option("timestampFormat", "dd-MM-yyyy")
      .schema(schema)
      .xml("test.xml").show()
  }
}
```

srowen:
A few things are going on here -

You'll need to specify `DateType` in the schema as this isn't a timestamp.
I also found that schema inference doesn't correctly infer date, and can fix that, but it isn't your issue here.
It works if you use ISO formatted dates like "2021-01-01". 

[...]

PR #521, opened 2021-02-18
Fix type inference for Dates; code cleanup

srowen:
(This is related to #519 but not its resolution)


src/main/scala/com/databricks/spark/xml/util/InferSchema.scala
```diff
@@ -119,6 +119,7 @@ private[xml] object InferSchema {
         case v if isDouble(v) => DoubleType
         case v if isBoolean(v) => BooleanType
         case v if isTimestamp(v) => TimestampType
+        case v if isDate(v) => DateType
```
srowen, line 122: This is really the fix; the rest is pretty much code cleanup

PR #522, opened 2021-02-18
Fix translation of types like xs:long to LongType, not IntegerType

srowen:


src/main/scala/com/databricks/spark/xml/util/XSDToSchema.scala
```diff
@@ -86,13 +86,7 @@ object XSDToSchema {
...
         val schemaType = simpleType.getContent match {
           case restriction: XmlSchemaSimpleTypeRestriction =>
-            val matchType =
-              if (restriction.getBaseTypeName == Constants.XSD_ANYSIMPLETYPE) {
-                simpleType.getQName
-              } else {
-                restriction.getBaseTypeName
-              }
-            matchType match {
+            simpleType.getQName match {
```
srowen, line 89: @seddonm1 this change seems to 'fix' the translation, which otherwise was coming up with xs:int when the schema says xs:long. This was from your original code, and I wasn't sure if there was another reason it has to be done this way? it's already known to be a simple type and seems like you can just get the QName in this case.

seddonm1: Sorry, I can't remember. At the time this was more a brute force reverse engineering process as the XSD spec is big.

srowen: Got it. I wouldn't want to break your use case - there are some tests but far from comprehensive. I'll try to think about why this would be needed, but I don't know XSDs that well (and at least I know it doesn't work in a simple case here). Thanks!

PR #539, opened 2021-06-07
Fix missing value tag crash

haavard:
**Before this PR:** the attached test would crash because we're not providing a valueTag in our struct schema. This is a regression from a previous version, where this was simply ignored.

**After this PR:** character element is ignored instead when there is no value tag.


srowen:
OK, I think I'm OK with that. I'd rather do something coherent than fail, although now it's going to drop some data without a column to put it in, but, that's on the caller.

PR #542, opened 2021-06-09
Return null for primitive types when value matches nullValue string

### LINKED ISSUE #540: NULL values not working for numeric fields
I have a question similar to issue #227.
I have tried to load following XML File using pyspark. The data contain a tag with an empty value for the numeric volume field.
I would like to this field to be filled with null, but I receive 0.0.

Here is my sample data:
```
<?xml version="1.0" encoding="UTF-8" ?>
<TEST>
<Header Name="X" SequenceNumber="123"/>
<T Number="1" Volume="20"/>
<T Number="2" Volume=""/>
</TEST>

```
I have tried following code
```
from pyspark.sql.types import StringType, LongType, DoubleType, StructType, StructField, ArrayType
from pyspark.sql.functions import explode

XMLSchema = ( StructType (
	[StructField("Header", StructType(
		[StructField("_Name", StringType(), True)
		,StructField("_SequenceNumber", LongType(), True)
        ]), True)
    ,StructField("T", ArrayType(StructType(
       [StructField("_Number", LongType(), True)
       ,StructField("_Volume", DoubleType(), True)
       ,StructField("_VALUE", DoubleType(), True)
       ]), True), True)
    ]))

df = ( spark.read.format('com.databricks.spark.xml')
                .option("rowTag", "TEST")
                .option("nullValue", "")
                .schema(XMLSchema)
                .load("mnt/test/Alfred/A.xml")
                .select("Header",explode("T"))
     )

df.printSchema()
df.show()

```

With following result:
```
root
 |-- Header: struct (nullable = true)
 |    |-- _Name: string (nullable = true)
 |    |-- _SequenceNumber: long (nullable = true)
 |-- col: struct (nullable = true)
 |    |-- _Number: long (nullable = true)
 |    |-- _Volume: double (nullable = true)
 |    |-- _VALUE: double (nullable = true)

+--------+---------------+
|  Header|            col|
+--------+---------------+
|{X, 123}|{1, 20.0, null}|
|{X, 123}| {2, 0.0, null}|
+--------+---------------+
```
If I simply omit the "nullValue" option in my code, the line is skipped completely.
How can I achieve that.
Expected result:
```
+--------+---------------+
|  Header|            col|
+--------+---------------+
|{X, 123}|{1, 20.0, null}|
|{X, 123}|{2, null, null}|
+--------+---------------+
```

srowen:
Closes #540 

While the code did handle detection of input matching the `nullValue` and returning null, in paths where the column type was a primitive in Java like `DoubleType`, the cast turned it into 0.0. This adds additional checking to perform the same check earlier.

### DIFF
--- a/src/main/scala/com/databricks/spark/xml/util/TypeCast.scala
+++ b/src/main/scala/com/databricks/spark/xml/util/TypeCast.scala
@@ -46,11 +46,9 @@ private[xml] object TypeCast {
   private[xml] def castTo(
       datum: String,
       castType: DataType,
-      options: XmlOptions,
-      nullable: Boolean = true): Any = {
-    if (datum == options.nullValue &&
-      nullable ||
-      (options.treatEmptyValuesAsNulls && datum == "")){
+      options: XmlOptions): Any = {
+    if ((datum == options.nullValue) ||
+        (options.treatEmptyValuesAsNulls && datum == "")) {
       null
     } else {
       castType match {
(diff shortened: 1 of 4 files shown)


src/main/scala/com/databricks/spark/xml/util/TypeCast.scala
```diff
@@ -46,11 +46,9 @@ private[xml] object TypeCast {
   private[xml] def castTo(
       datum: String,
       castType: DataType,
-      options: XmlOptions,
-      nullable: Boolean = true): Any = {
-    if (datum == options.nullValue &&
-      nullable ||
-      (options.treatEmptyValuesAsNulls && datum == "")){
+      options: XmlOptions): Any = {
+    if ((datum == options.nullValue) ||
```
srowen, line 50: `nullable` was not used - always `true`. The code path here doesn't actually care about nullability of the column, which is ... OK ? if it returns `null` where it's not allowed, because the input said it was null, that's an error later anyway and probably correctly.

PR #545, opened 2021-06-17
Change $default_branch to static branches (and set Java version correctly)

HyukjinKwon:
This PR proposes to use:
- static branches instead of `$default-branch` that's to be used as a template.
- use `adopt@1.8` syntax (as documented in https://github.com/olafurpg/setup-scala). Otherwise it picks JDK 11 for `java-version: '8'`.


srowen:
Nice, thank you!

PR #561, opened 2021-10-14
build: :sparkles: automatic developer environment provisioning with gitpod

edmondo1984:
This PR adds to spark-xml GitPod support and modify Github workflow actions so to run test on each branch.

### DIFF
--- a/.github/workflows/test_spark_2_java_8.yml
+++ b/.github/workflows/test_spark_2_java_8.yml
@@ -1,9 +1,5 @@
 name: Spark 2 / Java 8
-on:
-  push:
-    branches: [master]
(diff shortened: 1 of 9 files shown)


srowen:
I see - I recall I had trouble updating SBT in this project, but maybe it's resolvable now. That would be a fine thing to try to fix directly, separately.

Spark itself only supports Java 11, though 99% works on Java 17 already in 3.2.0. When that's out, then it's reasonable to test vs later Java versions. For now just use a lower version, yeah.

I think it's kind of rare to want to test vs different versions of Java for the occasional contributors, as the build matrix covers supported versions already

PR #594, opened 2022-07-06
Fix data leakage in exception message

### LINKED ISSUE #593: Data leakage in exception message
When Spark XML detects malformed input data, sensitive information contained in the record can be leaked into logs by the exception message in [StaxXmlParser.scala](https://github.com/databricks/spark-xml/blob/master/src/main/scala/com/databricks/spark/xml/parsers/StaxXmlParser.scala) lines 106-107:
```scala
throw new IllegalArgumentException(
  s"Malformed line in FAILFAST mode: ${abbreviatedRecord.replaceAll("\n", "")}", cause)
```

May we change this to the following?
```scala
logger.debug("Malformed line:", abbreviatedRecord.replaceAll("\n", ""))
throw new IllegalArgumentException("Malformed line in FAILFAST mode", cause)
```

Mike-Robert-CHC:
Resolves #593

### DIFF
--- a/src/main/scala/com/databricks/spark/xml/parsers/StaxXmlParser.scala
+++ b/src/main/scala/com/databricks/spark/xml/parsers/StaxXmlParser.scala
@@ -103,8 +103,8 @@ private[xml] object StaxXmlParser extends Serializable {
       case FailFastMode =>
         val abbreviatedRecord =
           if (record.length() > 1000) record.substring(0, 1000) + "..." else record
-        throw new IllegalArgumentException(
-          s"Malformed line in FAILFAST mode: ${abbreviatedRecord.replaceAll("\n", "")}", cause)
+        logger.debug("Malformed line:", abbreviatedRecord.replaceAll("\n", ""))
+        throw new IllegalArgumentException("Malformed line in FAILFAST mode", cause)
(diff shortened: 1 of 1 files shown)


src/main/scala/com/databricks/spark/xml/parsers/StaxXmlParser.scala
```diff
@@ -103,8 +103,8 @@ private[xml] object StaxXmlParser extends Serializable {
       case FailFastMode =>
         val abbreviatedRecord =
           if (record.length() > 1000) record.substring(0, 1000) + "..." else record
-        throw new IllegalArgumentException(
-          s"Malformed line in FAILFAST mode: ${abbreviatedRecord.replaceAll("\n", "")}", cause)
+        logger.debug("Malformed line:", abbreviatedRecord.replaceAll("\n", ""))
```
srowen, line 106: How about:
- Make this consistent with the DropMalformedMode below, logging the line
- But leave `cause` in the exception rather than log here
- Change both to INFO? makes more sense for DropMalformedMode than WARN, really

I'm hesitant to make it DEBUG as those logs typically aren't on and makes diagnosing something that already happened, maybe can't be easily reproduced, harder.

PR #596, opened 2022-07-10
[Clean Up] Remove some duplicated code with Spark and use the ones directly from Spark Repo

ericsun95:
The `src/main/scala/com/databricks/spark/xml/util/CompressionCodecs.scala` and `src/main/scala/com/databricks/spark/xml/util/ParseMode.scala` are duplicated with the corresponding classes in the `org.apache.spark.sql.catalyst.util`. Clean this up to make sure it is up-to-date.


src/main/scala/com/databricks/spark/xml/util/CompressionCodecs.scala
```diff
@@ -1,51 +0,0 @@
-/*
```
ericsun95, line 1: Duplicate of https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/CompressionCodecs.scala

src/main/scala/com/databricks/spark/xml/util/ParseMode.scala
```diff
@@ -1,59 +0,0 @@
-/*
```
ericsun95, line 1: Duplicate of https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/ParseMode.scala

srowen (commented):
I think this change is OK, needs to happen

src/main/scala/com/databricks/spark/xml/util/XmlFile.scala
```diff
@@ -15,21 +15,21 @@
...
 package com.databricks.spark.xml.util
 
-import java.io.CharArrayWriter
-import java.nio.charset.Charset
-import javax.xml.stream.XMLOutputFactory
-
-import scala.collection.Map
-
 import com.databricks.spark.xml.parsers.StaxXmlGenerator
+import com.databricks.spark.xml.{XmlInputFormat, XmlOptions}
```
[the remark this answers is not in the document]
ericsun95: Okay, sure. I can do that.

PR #615, opened 2022-12-20
Upgrade SBT

ganeshchand:
- Updated .gitignore to ignore files/folders specific to VS Code Scala Metals
- Updated build.properties
- Minor refactoring to build.sbt to conform to slash syntax

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,7 @@ target/
 .idea_modules/
 .bsp/
 .DS_Store
+.metals/
+.vscode/
+.bloop/
+metals.sbt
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -3,3 +3,7 @@ target/
 .idea_modules/
 .bsp/
 .DS_Store
+.metals/
```
srowen, line 6: Could you revert this?

srowen (changes requested):
Looks OK otherwise if tests pass. I seem to recall the error happened during release, but we'll find out

srowen (approved):
Nevermind it's fine

Issue #612, opened 2022-12-19
XML Timestamp parsing without timezone


JorisTruong:
# Description

Currently encountering some issue when parsing with a specific timestamp format. I have XML data with timestamp of this format: `yyyy/MM/dd HH:mm:ss`, and I am trying to read it using the `timestampFormat` option.

Looking at the `parseXmlTimestamp()` function in [TypeCast.scala](https://github.com/databricks/spark-xml/blob/master/src/main/scala/com/databricks/spark/xml/util/TypeCast.scala), it seems that `spark-xml` will be using `Timestamp.from(ZonedDateTime.parse(value, format).toInstant)` to parse the timestamp. However, with a format like `yyyy/MM/dd HH:mm:ss`, there is no timezone, which causes the function to fail:
![image](https://user-images.githubusercontent.com/29867169/208391764-8ddd1af1-a122-45b7-b204-fe130004fd76.png)
It will then return null values for the whole timestamp column.

# To reproduce
`time.xml` file:
```
<book>
    <author>John Smith</author>
    <time>2011-12-03T10:15:30Z</time>
    <time2>2011/03/12 10:15:30</time2>
</book>
```

PySpark code to read:
```
df = spark.read \
        .format("com.databricks.spark.xml") \
        .options(rowTag="book", timestampFormat="yyyy/MM/dd HH:mm:ss") \
        .schema("author STRING, time TIMESTAMP, time2 TIMESTAMP") \
        .load("<path_to_file>")

display(df)
df.printSchema()
```

Result:
|author|time|time2|
|-------|-----|------|
|John Smith|2011-12-03T10:15:30.000+0000|null|

# Suggestion
We may want to add a default timezone in the `parseXmlTimestamp()` function, like this:
![image](https://user-images.githubusercontent.com/29867169/208391973-518872fe-f1be-4238-8d60-a4e6452c77db.png)

Any thoughts about adding a `timeZone` option?
```
private[xml] class XmlOptions(
    @transient private val parameters: Map[String, String])
  extends Serializable {

  def this() = this(Map.empty)

  val charset = parameters.getOrElse("charset", XmlOptions.DEFAULT_CHARSET)
  ...
  val timestampFormat = parameters.get("timestampFormat")
  val timeZone = parameters.getOrElse("timeZone", XmlOptions.DEFAULT_TIME_ZONE) // <----- HERE
  val dateFormat = parameters.get("dateFormat")
}

private[xml] object XmlOptions {
[...]

PR #616, opened 2022-12-21
feat: added timeZone option

JorisTruong:
#612

### DIFF
--- a/src/main/scala/com/databricks/spark/xml/XmlOptions.scala
+++ b/src/main/scala/com/databricks/spark/xml/XmlOptions.scala
@@ -64,6 +64,7 @@ private[xml] class XmlOptions(
     parameters.getOrElse("wildcardColName", XmlOptions.DEFAULT_WILDCARD_COL_NAME)
   val ignoreNamespace = parameters.get("ignoreNamespace").map(_.toBoolean).getOrElse(false)
   val timestampFormat = parameters.get("timestampFormat")
+  val timeZone = parameters.getOrElse("timeZone", XmlOptions.DEFAULT_TIME_ZONE)
   val dateFormat = parameters.get("dateFormat")
 }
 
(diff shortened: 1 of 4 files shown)


src/main/scala/com/databricks/spark/xml/XmlOptions.scala
```diff
@@ -64,6 +64,7 @@ private[xml] class XmlOptions(
     parameters.getOrElse("wildcardColName", XmlOptions.DEFAULT_WILDCARD_COL_NAME)
   val ignoreNamespace = parameters.get("ignoreNamespace").map(_.toBoolean).getOrElse(false)
   val timestampFormat = parameters.get("timestampFormat")
+  val timeZone = parameters.getOrElse("timeZone", XmlOptions.DEFAULT_TIME_ZONE)
```
srowen, line 67: Instead of exposing an option, use the value of Spark conf `spark.sql.session.timeZone` below. We would need to document this behavior in the README

src/main/scala/com/databricks/spark/xml/util/TypeCast.scala
```diff
@@ -115,11 +115,14 @@ private[xml] object TypeCast {
...
 
   private def parseXmlTimestamp(value: String, options: XmlOptions): Timestamp = {
+    val timeZone = options.timeZone
     val formatters = options.timestampFormat.map(DateTimeFormatter.ofPattern).
       map(supportedXmlTimestampFormatters :+ _).getOrElse(supportedXmlTimestampFormatters)
     formatters.foreach { format =>
       try {
-        return Timestamp.from(ZonedDateTime.parse(value, format).toInstant)
+        return Timestamp.from(
+          ZonedDateTime.parse(value, format.withZone(ZoneId.of(timeZone))).toInstant
```
srowen, line 124: The new logic should only be applied to the custom timestamp format, not the built-in ones.
I suppose, if possible, you check to see if the pattern has a TZ and augment it with the TZ specified in `spark.sql.session.timeZone` if not.

srowen (changes requested):

PR #621, opened 2022-12-26
Allow custom timestamp with Spark timezone property

JorisTruong:
Related to issue #612 and to previous pull request #616.

There are still some issues as `spark.sql.session.timeZone` uses Java's `TimeZone.getDefault.getID` according to the source code **[here](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala#L2555)**, and it can result in a null value.

As a result, it will be mandatory to set `spark.sql.session.timeZone`, otherwise `spark-xml` will throw an `NoSuchElementException` when trying to retrieve the Spark property with `spark.conf.get()` method. Can reproduce this when running the `XmlPartitioningSuite`.

We may still need a default value for the timezone.


src/main/scala/com/databricks/spark/xml/DefaultSource.scala
```diff
@@ -70,7 +70,8 @@ class DefaultSource
     XmlRelation(
       () => XmlFile.withCharset(sqlContext.sparkContext, path, charset, rowTag),
       Some(path),
-      parameters,
+      parameters + ("timezone" ->
```
srowen, line 73: No need to pass this through. Just store it in TypeCast as a member

README.md
```diff
@@ -65,7 +65,7 @@ Defaults to `false`. New in 0.11.0.
 * `timestampFormat`: Specifies an additional timestamp format that will be tried when parsing values as `TimestampType` 
 columns. The format is specified as described in [DateTimeFormatter](https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html).
 Defaults to try several formats, including [ISO_INSTANT](https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html#ISO_INSTANT),
-including variations with offset timezones or no timezone (defaults to UTC). New in 0.12.0.  
+including variations with offset timezones or no timezone (defaults to UTC). If a custom format is used, it will check if it has a timezone in the pattern. If not, it will uses the default spark timezone from `spark.sql.session.timeZone`. New in 0.12.0.
```
srowen, line 68: Let's say "As of 0.16.0, if a custom format pattern is used without a timezone, the default Spark timezone specified by `spark.sql.session.timeZone` will be used" and similarly below

src/main/scala/com/databricks/spark/xml/util/TypeCast.scala
```diff
@@ -119,7 +119,15 @@ private[xml] object TypeCast {
       map(supportedXmlTimestampFormatters :+ _).getOrElse(supportedXmlTimestampFormatters)
     formatters.foreach { format =>
       try {
-        return Timestamp.from(ZonedDateTime.parse(value, format).toInstant)
+        // If format is not in supported and no timezone in format, use default Spark timezone
+        if (!supportedXmlTimestampFormatters.contains(format) && Option(format.getZone).isEmpty) {
```
srowen, line 123: Rather than do this, just break this method into two checks - one loop over built-in formats, then the custom format with special TZ handling

src/main/scala/com/databricks/spark/xml/util/TypeCast.scala
```diff
@@ -119,7 +119,15 @@ private[xml] object TypeCast {
...
+          return Timestamp.from(
+            ZonedDateTime.parse(value, format.withZone(ZoneId.of(options.timezone.get))).toInstant
```
srowen, line 125: Need to deal with timezone not being specified - I suppose throw an error explicitly?

srowen (changes requested):

src/main/scala/com/databricks/spark/xml/DefaultSource.scala
JorisTruong: I think I need a SparkSession to get `spark.sql.session.timeZone`. Is that ok to pass a SparkSession to TypeCast?

srowen: I'd just use SparkSession.getActiveSession() to get the active session where you need it, no need to pass it

JorisTruong: I tried using this as well as SparkSession.getDefaultSession(), but it kept returning me null. I believe this is because the code is running on executors: https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/SparkSession.scala#L1073

srowen: Oh I see, hm. OK, maybe this approach is the cleanest.

src/main/scala/com/databricks/spark/xml/util/TypeCast.scala
JorisTruong: Do you mean we should use two methods? If I understand correctly,

```
    val formatters = options.timestampFormat.map(DateTimeFormatter.ofPattern).
      map(supportedXmlTimestampFormatters :+ _).getOrElse(supportedXmlTimestampFormatters)
    formatters.foreach { format =>
```

This block of code means that the current `parseXmlTimestamp()` method already loops over built-in formats + the custom format. It is just that if the specified format does not contains a timezone, the custom format does not work. So why break it if the original idea is to loop over all formats?

If you meant just break in the method, doesn't the if branch does it?

src/main/scala/com/databricks/spark/xml/util/TypeCast.scala
JorisTruong: With the current approach to put the timezone in `XmlRelation` of  `createRelation()`, isn't it better the create a default timezone in case `spark.sql.session.timeZone` is not set?

srowen: Hm, maybe, I just feel like that's an error. You are parsing times that are ambiguous without a timzeone, and didn't give a timezone - just assuming "UTC" or something doesn't quite seem right vs highlighting the error.

src/main/scala/com/databricks/spark/xml/util/TypeCast.scala
srowen: It's all about the same, but why combine two types of input that need different processing and then re-differentiate in a loop? handle the loop, then handle the special case. I think it's more straightforward, maybe not.

src/main/scala/com/databricks/spark/xml/util/TypeCast.scala
```diff
@@ -116,11 +116,32 @@ private[xml] object TypeCast {
...
+        )
+      } catch {
+        case _: Exception => // continue
+      }
       try {
-        return Timestamp.from(ZonedDateTime.parse(value, format).toInstant)
+        // Custom format without timezone or offset
+        return Timestamp.from(
+          ZonedDateTime.parse(value, format.withZone(ZoneId.of(options.timezone.get))).toInstant
+        )
```
JorisTruong, line 144: We can check if there is a timezone but there is no method to determine if an offset is defined

srowen: Hm. What if we parse, and if it fails with the error you are facing, add 'withZone'? the only problem is that may be a huge amount of perf overhead, so we'd have to have some way to test and store the format with the zone, only if it's needed. Not pretty but that would be the way forward

JorisTruong: When parsing fails, it will always throw a `DateTimeParseException`, no matter if it is because of the problem we are talking about or not. Rather than parsing the error message, I wrote a `isParseableAsZonedDateTime()` method and used it during the loop.

src/main/scala/com/databricks/spark/xml/DefaultSource.scala
```diff
@@ -67,11 +67,23 @@ class DefaultSource
...
-    XmlRelation(
-      () => XmlFile.withCharset(sqlContext.sparkContext, path, charset, rowTag),
-      Some(path),
-      parameters,
-      schema)(sqlContext)
+    val sparkTimezone = sqlContext.sparkContext.getConf.getOption(
+      "spark.sql.session.timeZone"
+    )
+
+    if (sparkTimezone.isDefined) {
```
srowen, line 74: Can you just vary `parameters` based on `sparkTimezone` rather than repeat the whole block?

src/main/scala/com/databricks/spark/xml/util/TypeCast.scala
```diff
@@ -112,15 +113,23 @@ private[xml] object TypeCast {
     // 2002-05-30T21:46:54+06:00
     DateTimeFormatter.ISO_OFFSET_DATE_TIME,
     // 2002-05-30T21:46:54.1234Z
-    DateTimeFormatter.ISO_INSTANT
+    DateTimeFormatter.ISO_INSTANT.withZone(ZoneId.of("UTC"))
```
srowen, line 116: This should already be UTC; I don't think we want to change this
https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html#ISO_INSTANT

src/main/scala/com/databricks/spark/xml/util/TypeCast.scala
```diff
@@ -112,15 +113,23 @@ private[xml] object TypeCast {
...
   )
 
   private def parseXmlTimestamp(value: String, options: XmlOptions): Timestamp = {
     val formatters = options.timestampFormat.map(DateTimeFormatter.ofPattern).
       map(supportedXmlTimestampFormatters :+ _).getOrElse(supportedXmlTimestampFormatters)
     formatters.foreach { format =>
       try {
-        return Timestamp.from(ZonedDateTime.parse(value, format).toInstant)
+        if (isParseableAsZonedDateTime(value, format)) {
```
srowen, line 124: Can you vary `format` only based on this check? to avoid repeating

src/main/scala/com/databricks/spark/xml/util/TypeCast.scala
```diff
@@ -268,4 +277,17 @@ private[xml] object TypeCast {
       TypeCast.castTo(data, FloatType, options).asInstanceOf[Float]
     }
   }
+
+  private[xml] def isParseableAsZonedDateTime(value: String,
```
srowen, line 281: For later, we might have to figure out a way to cache the parsing of a custom format into a pattern, and maybe this check, because it'll happen for every single row

srowen (changes requested):

src/main/scala/com/databricks/spark/xml/util/TypeCast.scala
JorisTruong: I haven't been able to run `Timestamp.from(ZonedDateTime.parse(value, DateTimeFormatter.ISO_INSTANT).toInstant)`. The doc also says:

> As such, an Instant cannot be formatted as a date or time without providing some form of time-zone.

I feel like that ISO_INSTANT should not be in `supportedXmlTimestampFormatters`.
Same issue here: https://stackoverflow.com/questions/25612129/java-8-datetimeformatter-and-iso-instant-issues-with-zoneddatetime

src/main/scala/com/databricks/spark/xml/util/TypeCast.scala
JorisTruong: I should remove the parsing error check, as it should be caught in `parseXmlTimestamp()` method

src/main/scala/com/databricks/spark/xml/util/TypeCast.scala
srowen: I think we need to support the format; I think I copied these from a list of standard formats according to XSD specs or something. It's super standard.

We're parsing rather than formatting here. What's the issue - does it not work with ZonedDateTime? maybe never did, if so.

This would probably also be better if we somehow cached the parsing of these built-in patterns as parser/formatters, but that can happen separately with some optimization.

srowen: Hm OK I see why you made that change, it doesn't seem to think it's "UTC" otherwise, when it should be by nature. OK leave it in

srowen: OK different idea - what if we write the parsing without ZonedDateTime? `Timestamp.from(Instant.from(format.parse(value)))` Does that help? seems to be simpler, not sure why it wasn't written that way in the first place

JorisTruong: No, ISO_INSTANT does not work with ZonedDateTime

srowen:
Take a look at this change -- I think the core of this works? maybe adapt this approach
https://github.com/databricks/spark-xml/pull/624

JorisTruong:
I think you have the best answer; I added some more tests in the pull request. I'll try to look into why tests are failng though

srowen:
I think I figured out the test failure - tiny but subtle issue in handling the param map. See my latest push

src/main/scala/com/databricks/spark/xml/util/TypeCast.scala
```diff
@@ -116,11 +116,28 @@ private[xml] object TypeCast {
...
+    supportedXmlTimestampFormatters.foreach { format =>
+      try {
+        return Timestamp.from(Instant.from(format.parse(value)))
+      } catch {
+        case _: Exception => // continue
+      }
+    }
+    options.timestampFormat.foreach { formatString =>
+      // Check if there is offset or timezone and apply Spark timeZone if not
+      val hasTemporalInformation = formatString.indexOf("V") +
```
srowen, line 128: I'm not sure we need this - I found that the docs for "withZone" say that it's ignored if the pattern contains timezone info. So it seems like it will just be a default. OK to write a test for that though!

srowen (commented):
Looking good to me otherwise

src/main/scala/com/databricks/spark/xml/util/TypeCast.scala
JorisTruong: Yes I also saw this, but then I encountered some problems between Java 8 and Java 11. We can see in Java 8 [here](https://github.com/frohoff/jdk8u-dev-jdk/blob/master/src/share/classes/java/time/format/Parsed.java#L586) that zone is prioritized over offset, but in Java 9 [here](https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/master/src/java.base/share/classes/java/time/format/Parsed.java#L595), offset is prioritized over zone. I made an example below to see the differences:
- Java 8: image
- Java 11: image

So I don't think we can always apply `withZone()`, and that's why I wanted to use `hasTemporalInformation`. unfortunately, I haven't been able to think of a cleaner way to do this

srowen: OK, let's leave it this way for now. If you have a sec, throw in a comment about the Java version

srowen (approved):

PR #629, opened 2023-02-01
Initial pass at supports 'paths' data source option with multiple file paths

srowen:
(Pull request for testing - TBD)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2,7 +2,7 @@ import com.typesafe.tools.mima.core.MissingClassProblem
 
 name := "spark-xml"
 
-version := "0.16.0"
+version := "0.17.0"
 
 organization := "com.databricks"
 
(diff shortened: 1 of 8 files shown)


zsxwing:
Is it possible to write a test using `spark.read.format("xml").load(paths: _*)` to verify the change?

srowen:
@zsxwing heh yeah it doesn't work. Looking more closely, it seems like "paths" is a DSv2 thing, and only "path" is set for legacy V1 sources (like this). I can try a hack but really this seems tied up with V2 support I guess, and I don't know how to do that

zsxwing:
After checking the code, I also realized that `load(paths)` in DSv1 only works for `FileFormat` which is the internal interface for built-in file formats 😢 

Thanks for quickly working on this! I think we have to give up this as moving DSv2 is unrealistic.

srowen:
OK canning this for now. Thanks for checking!

Issue #554, opened 2021-08-19
Adding complexContent Support for XsdToSchema


exandi:
Hey Community,
is there any thing in the pipeline to support the XSD-Element `complexContent`

Currently I am planning to read an XSD File to an Spark DF Schema, but if the XSD contains an `complexContent` Field the conversion fails.

srowen:
I don't plan to work on this, but you are welcome to propose a PR if it's clean and adds support for this. 
But do you mean XSD validation, or XSD to schema? complexContent probably won't translate to a Spark schema.
For XSD validation it probably already works? because it's just applying a standard XML parser with the XSD to it.

exandi:
I tried to convert an XSD File to a Schema, will provide tomorrow the details, but if I comment the compexContent out it works fine, else it crashes. I will also provide the stacktrace tomorrow.

exandi:
Hey,
sorry for the delay.

I use a kerborised hadoopcluster (HDP 3.1.4) with spark `2.3.2.3.1.4.0-315` To get a Spark context I use Jupyter Notebook with the following command:

```
import os, sys
os.environ["HADOOP_CONF_DIR"] = "/etc/hadoop_spark/conf"
os.environ["SPARK_HOME"] = "/usr/hdp/current/spark2-client"
os.environ["PYSPARK_PYTHON"] = os.environ["HOME"] + "/nfs/bin/python"
os.environ['PYSPARK_SUBMIT_ARGS'] = f'--jars /jars/spark-xml_2.11-0.12.0.jar","/jars/xmlschema-core-2.2.5.jar pyspark-shell'
```

I need to add the `xmlschema-core.jar` as well. Else I get a ClassNotFound Exception.

After I created a Spark Context via:
```
spark = SparkSession \
    .builder \
    .appName("XMLTest") \
    .enableHiveSupport() \
    .config("spark.driver.allowMultipleContexts", "true") \
    .getOrCreate()
```

I am able to call the XSDToSchema read Method via:
`print(str(spark._jvm.com.databricks.spark.xml.util.XSDToSchema.read(xsdasstring)))` where xsdasstring is the XSD-file read in.

the XSD-File looks like:

> <?xml version="1.0" encoding="UTF-8"?>
> <xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
>            elementFormDefault="qualified">
> 
> <xs:element name="employee" type="fullpersoninfo"/>
> 
> <xs:complexType name="personinfo">
[...]

PR #631, opened 2023-02-12
Parse complexContent with extension element

shuch3ng:
Resolves an issue reported in https://github.com/databricks/spark-xml/issues/554

- adding the capability to parse extension element in complexContent


srowen (approved):
OK I buy it, looks plausible

PR #660, opened 2023-09-02
Shortcut common type inference cases to fail fast, speed up inference

srowen:
In schema inference, many different types are tried out for each input. This can get really slow in some cases, especially where the true type is just 'string'. This adds several shortcuts in the type inference code, to fail fast before expensive parsing code is run, where it's clear the parsing won't work. This also avoids using a thrown exception in one case for better speed.


HyukjinKwon (approved):

srowen:
I've got a customer checking out this change too. If I put it in, I'll also need to get this applied to the oustanding patch vs Spark that ports this.

Issue #664, opened 2023-10-09
Generated files does not have .xml extension


dolfinus:
Hi.

I've created simple dataframe:
```python
from pyspark.sql import SparkSession
from pyspark.sql.types import StructType, StructField, TimestampType
from datetime import datetime, timezone

spark = SparkSession.builder.config("spark.jars.packages", "com.databricks:spark-xml_2.12:0.17.0").getOrCreate()
schema = StructType([StructField("created-at", TimestampType())])

df = spark.createDataFrame([{"created-at": datetime.now(tz=timezone.utc)}], schema=schema)
df.show(10, False)

df.write.format("xml").option("timestampFormat", "yyyy-MM-dd HH:mm:ss.SSSXXX").mode("overwrite").save("2.xml")
```
|created-at                |
|--------------------------|
|2023-10-09 09:05:24.269352|

Then saved it as `xml`:
```python
df.repartition(1).write \
  .format("xml") \
  .mode("overwrite") \
  .option("compression", None) \
  .option("rowTag", "item") \
  .save("2.xml")
```

This is content of `2.xml` folder:
```bash
> ls -la 2.xml
drwxr-xr-x  2 maxim maxim   84 окт  9 09:18 ./
drwxr-xr-x 19 maxim maxim 4096 окт  9 09:18 ../
-rw-r--r--  1 maxim maxim  156 окт  9 09:18 part-00000
-rw-r--r--  1 maxim maxim   12 окт  9 09:18 .part-00000.crc
-rw-r--r--  1 maxim maxim    0 окт  9 09:18 _SUCCESS
-rw-r--r--  1 maxim maxim    8 окт  9 09:18 ._SUCCESS.crc
```

File `2.xml/part-00000` has the following content:
```xml
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<ROWS>
    <item>
        <created-at>2023-10-09T09:05:24.269352Z</created-at>
    </item>
</ROWS>
```

But it does not have `.xml` extension. Is that an expected behavior?

srowen:
It's expected. I don't know of a way to control this, and won't change it at this point (the library is now in Spark anyway)

dolfinus:
I see, `rdd.saveAsTextFile` creates directory with files without extensions. I think it is worth mentioning in Readme.

Issue #662, opened 2023-10-09
Timestamps not matching format are replaced with nulls


dolfinus:
Hi.

I'm trying to parse simple xml file:
```xml
<item>
  <created-at>2021-01-01T01:01:01+00:00</created-at>
</item>
```

```python
from pyspark.sql import SparkSession
from pyspark.sql.types import StructType, StructField, TimestampType

spark = SparkSession.builder.config("spark.jars.packages", "com.databricks:spark-xml_2.12:0.17.0").getOrCreate()
schema = StructType([StructField("created-at", TimestampType())])
spark.read.format("xml").options(rowTag='item').schema(schema).load("1.xml").show()
```

Result:
|         created-at|
|-------------------|
|2021-01-01 01:01:01|

But if timestamp does not match format, e.g. `T` is replaced with space:
```xml
<item>
  <created-at>2021-01-01 01:01:01+00:00</created-at>
</item>
```

It is read as `null`:
|created-at|
|----------|
|      null|

I see that there is an option `mode` with `PERMISSIVE` as default, which leads to `when it encounters a field of the wrong datatype, it sets the offending field to null`. But malformed value is not being added to column `_corrupt_record` because there is nothing wrong with xml structure.
So there is no way to detect if input file contains tag with wrong field value or `nullValue`, unless user set a different `mode`.
Is that desired behavior?

dolfinus:
Tried:
```python
from pyspark.sql import SparkSession
from pyspark.sql.types import StructType, StructField, TimestampType, StringType

spark = SparkSession.builder.config("spark.jars.packages", "com.databricks:spark-xml_2.12:0.17.0").getOrCreate()
schema = StructType([StructField("created-at", TimestampType()), StructField("_corrupt_record", StringType())])
spark.read.format("xml").options(rowTag='item').schema(schema).load("1.xml").show(10, False)
```

```
|created-at|_corrupt_record                                                      |
|----------|---------------------------------------------------------------------|
|null      |<item>\n  <created-at>2021-01-01 01:01:01+00:00</created-at>\n</item>|
```

It is worth mentioning that `_corrupt_record` should be explicitly added to dataframe schema.

PR #674, opened 2023-12-16
Add notes about file extensions and _corrupt_record to documentation

dolfinus:
In #662 and #664 I faced unexpected behavior which should be mentioned in documentation.


README.md
```diff
@@ -46,6 +46,13 @@ When reading files the API accepts several options:
   * `FAILFAST` : throws an exception when it meets corrupted records.
 * `inferSchema`: if `true`, attempts to infer an appropriate type for each resulting DataFrame column, like a boolean, numeric or date type. If `false`, all resulting columns are of string type. Default is `true`.
 * `columnNameOfCorruptRecord`: The name of new field where malformed strings are stored. Default is `_corrupt_record`.
+
+  Note: you should explicitly add `_corrupt_record` field to dataframe schema, like this:
```
srowen, line 50: This isn't accurate for the common case though. It's added automatically when you infer the schema.
I think the error is pretty clear here, in the case you specify the schema, don't add this, and it tells you it expected it.

dolfinus: Ok, it was not so obvious for me, so I added a note

srowen (approved):

